Issue #463, opened 2016-12-25
Left bleed example is bleeding


mayagao:
<img width="657" alt="screen shot 2016-12-25 at 11 38 42 am" src="https://cloud.githubusercontent.com/assets/5186564/21472688/dfab186a-ca99-11e6-96c6-6c1770442f5f.png">

mayagao:
This is fixed!

PR #501, opened 2017-01-03
improve bleed and float example

### LINKED ISSUE #499: Bleed classes don't "meet the edge of the screen"
The current description for bleed classes says:

> Bleed classes allow an element inside a centered container to escape the container and meet the edge of the screen

And right below that is an example where the element does not in fact reach the edge of the screen. What is it reaching to the edge of? What's really going on here?

mayagao:
screen shot 2017-01-03 at 6 17 09 pm
screen shot 2017-01-03 at 6 17 04 pm

closes #499 #463

@davidtheclark @samanpwbb review

### DIFF
--- a/src/layout.css
+++ b/src/layout.css
@@ -1996,7 +1996,12 @@
  * When using the bleed classes, consider adding the `.clip` class on a parent element to horizontal overflow.
  * @group
  * @example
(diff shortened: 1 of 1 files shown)


samanpwbb:
Much improved, thanks @mayagao

PR #503, opened 2017-01-03
Magic focus ring management

### LINKED ISSUE #493: Only show focus rings when tabbing
Palantir has a neat, simple little trick (see http://blueprintjs.com/docs/#components.button): when you click around the page, there are no focus outlines; but when you press tab to start keyboard-controlling, they appear. Then they disappear again when you click.

They are toggling a class on `<body>` that determines whether these outlines show up or not. This requires some JS —— but *we* can add similar JS unobtrusively since we are already adding some JS to load SVGs. As long as we can keep it small and vanilla, it should be fine. (Though we then should rename `assembly-svg.js` to `assembly.js`.)

@mayagao would be very happy with this change. I'm sure @dasulit and @tristen would as well. So maybe it's worth a shot.

davidtheclark:
Closes #493.

I had to modify the build process some for this. Instead of `assembly-svg.js`, we now have `assembly.js`, which concatenates the SVG-loading JS with any scripts in `src/js`. I also added uglification and silly optimization (via [optimize-js](https://github.com/nolanlawson/optimize-js)) to that concatenated script.

`focus-ring.js` is the simple little script that manages the focus ring.

I added `focus.css`, where the focus management nastiness is congregated and contained. After `focus-ring.js` runs, nothing gets its browser-default focus outline. Instead, once you press tab our special focus outline is added to everything. I made it a bright, mostly opaque blue. If anybody wants to tinker with that, it's the `--focus-shadow` variable.

This seems to be working nicely!

@samanpwbb and @mayagao for review. If you like how this works, I'll add some documentation to the homepage before merging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,13 +9,13 @@
     "check-size": "node scripts/check-size",
     "build:site": "node scripts/build-site",
     "build:css": "node scripts/build-css",
-    "build:svg": "node scripts/build-svg-sprite",
-    "build": "rm -rf dist && npm-run-all --parallel build:css build:svg build:site",
+    "build:js": "node scripts/build-js",
+    "build": "rm -rf dist && npm-run-all --parallel build:css build:js build:site",
     "lint:css": "stylelint src/*.css",
     "lint:js": "eslint .",
     "lint": "npm run lint:css && npm run lint:js",
     "serve": "node scripts/serve",
-    "start": "npm run build:svg && npm run serve",
+    "start": "npm run build:js && npm run serve",
     "test:jest": "jest",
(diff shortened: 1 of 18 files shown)


samanpwbb (approved):
looks perfect to me.

davidtheclark:
@mayagao I'll merge this now but welcome your feedback anytime you're able to get around to it.

mayagao:
Thank you for doing this! 

I would tone down the shadow even more to 0.25. It looks weird for elements like radio buttons and checkboxes. But I guess that's fine here.

PR #504, opened 2017-01-03
Rename align classes and fix vertical align example

### LINKED ISSUE #496: Maybe we should rename .txt-align-* to .inline-align-*
The reason is that these classes are for layout, not just typography. They are equally useful for aligning `inline-block`s of all kinds.

davidtheclark:
Closes #496.

@dasulit for review.


dasulit (approved):
Looks great!

PR #507, opened 2017-01-03
Async-defer the JS

davidtheclark:
We should be able to safely recommend that people use `async` and `defer` attributes when they load the Assembly JS. They could also put it at the bottom of the `<body>`. This PR documents that and uses the documentation pages as proof-of-concept.

@tristen for review.

### DIFF
--- a/scripts/build-js.js
+++ b/scripts/build-js.js
@@ -21,7 +21,8 @@ function buildJs(outfile) {
     concatJs()
   ])
     .then((data) => {
-      return UglifyJS.minify(data.join(''), { fromString: true }).code;
+      const allJs = data.join('');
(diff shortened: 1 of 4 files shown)


tristen:
Docs look good but `async` appears to not render icons:

screen shot 2017-01-03 at 1 35 17 pm

`defer` on it's own works. Just not the combination of `async` or on it's own.

davidtheclark:
Egad. I've seen it work a couple of times but now I'm also seeing it not work. Gotta fix.

davidtheclark:
@tristen I fixed the SVG loading script a little. The problem was that the `DOMContentLoaded` listener could be added after the event had already fired. So we're now checking `readyState`, and if it's already not `loading` add the SVGs; otherwise add the listener.

Issue #513, opened 2017-01-03
replace use of `var(--blue)` for defaults with `var(--default)`


samanpwbb:
someone might want a custom build where the default color isn't blue

davidtheclark:
I can take care of this ...

PR #515, opened 2017-01-03
Use the word default

davidtheclark:
- Uses `default-interactive-color` and `default-interactive-color-dark` variables throughout, instead of `blue` and `blue-dark`, closing #513.
- Changes "universal" to "default" in color variants. A breaking change for the Mapbox custom build.

@samanpwbb for review.


samanpwbb:
looks great

PR #516, opened 2017-01-03
Add txt- prefix

davidtheclark:
Adds `txt-` prefix to
- `uppercase`
- `lowercase`
- `capitalize`
- `capitalize-first`
- `em`
- `code` (I thought this was appropriate because it's really only for code-within-text)

Tries to fix the places those classes were used in the site.

Moves `align-*` classes lower down in typography utils because they stand out for not having a `txt-` prefix.

Also fixes a small unrelated thing: ensures that JS is built during the `build-site` script that runs on CI.

@samanpwbb for review.

### DIFF
--- a/scripts/build-site.js
+++ b/scripts/build-site.js
@@ -4,11 +4,15 @@
 require('babel-register');
 const renderSite = require('./render-site');
 const buildCss = require('./build-css');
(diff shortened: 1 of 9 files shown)


samanpwbb (approved):
hooray

PR #468, opened 2016-12-25
some style changes 

mayagao:
address some of the issues i created

PR #517, opened 2017-01-04
adjust style of documentation page and nav bar, closes #466

### LINKED ISSUE #466: Blue background is slightly hard to read
white text on bright blue background is a bit hard to read

screen shot 2016-12-25 at 11 12 15 am

samanpwbb:
Goals:

- Clearer hierarchy in sidebar (before, main page links and subsections of documentation looked too similar)
- More subtle sidebar (it has too much visual weight in live design)
- Less messy example display (the dotted border was good in theory but just using Maya's design from her PR looks and feels better).
- More legible class lists – they should stand out and each item should be clearly separate from other items.

Looks like this:

screen shot 2017-01-03 at 7 35 33 pm

@mayagao and @dasulit for review


site/navigation.js
```diff
@@ -29,7 +26,12 @@ class Navigation extends React.Component {
               className={linkClasses}
               href={member.route}
             >
+              {level === 0 ? <svg
+                style={{ marginLeft: '-18px' }}
+                className='absolute left icon'
+                dangerouslySetInnerHTML={{ __html: '<use xlink:href="#icon-caret-right"></use>' }} /> : ''}
```
dasulit: Breaking up the nav like this is a great improvement, but the caret definitely reads as an expandable dropdown to me. Can you try experimenting with a `ul` or a dash or something?

dasulit (changes requested):
I'm very excited about the changes in here; sidebar looks awesome, and I love Maya's example template! One suggestion about rethinking carets as bullet points, otherwise this is feeling so complete

mayagao:
1. class name is the most important info, but it's using the smallest font size, the purple bg also makes it look clickable:

screen shot 2017-01-04 at 2 13 52 pm

in https://github.com/mapbox/assembly/pull/468:

screen shot 2017-01-04 at 2 15 59 pm


2. something like this for sidebar since hierarchy between classnames are more important. 

screen shot 2017-01-04 at 2 11 10 pm

or even drop subsections from the sidebar and use a top nav layout for non-documentation pages

screen shot 2017-01-04 at 2 30 24 pm

samanpwbb:
> class name is the most important info, but it's using the smallest font size, the purple bg also makes it look clickable:

Font size gets small on small screens to prevent single rules from on line-breaking. I can look into another technique for doing the same thing. As for the backgrounds - without them, one class visually blends with the previous class and it's hard to tell what the actual rule is that is being documented. In CSS, commas are valid syntax. I can try something more subtle.

> something like this for sidebar since hierarchy between classnames are more important.

I can try for something closer to that

mayagao (approved):

samanpwbb:
I fixed all the problems! Merging.

PR #526, opened 2017-01-04
add missing gutter rules, to match margin/padding

samanpwbb:


davidtheclark:
I feel bad about adding the `hmin3` and similar, since it's pretty much a useless class. I guess it's the same old pragmatic-vs-consistent question, inescapable with sizing and spacing utilities, that is going to continue to plague this framework until sweet death takes it. I guess let's merge.

davidtheclark (approved):

PR #543, opened 2017-01-04
add logo

samanpwbb:
screen shot 2017-01-04 at 12 25 05 pm

@davidtheclark for review wrt the implementation


davidtheclark:
Did you try making this SVG a React component? If that works, it might be a little less hacky.

PR #544, opened 2017-01-04
Version assembly

### LINKED ISSUE #521: Versioning
Let's think about how this will be versioned. 

Right now we are recommending that people use `mapbox.com/assembly/assembly.css`. That will be the version of the stylesheet on the live site. But if there are breaking changes, then, other projects will break without warning or a chance to opt-out of the update. So I don't think it's a good system.

We could suggest people use [unpkg](https://unpkg.com/#/) and publish the `dist/` built files as part of the npm package (though not committed to the repo).

If we want a Mapboxier method for deployment, we should deploy versions to S3, right? Then serve them from there.

We also need a changelog.

tristen:
By copying the assets from `dist/` over to a directory defined by the version key in `package.json` and uploading the result to s3.

Closes #521 

---

### todo

- [x] I should probably use Promises instead of d3-queue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,5 +131,15 @@ npm start # Builds everything, starts a dev server, rebuilds & reloads on change
 
 npm run build:svg # Build SVGs
 ```
+### Releasing
+
+- Document changes in the [`CHANGELOG`](https://github.com/mapbox/assembly/blob/mb-pages/CHANGELOG.md).
+- Tag the version in git and `git push --tags`.
+- Increment the version key in [`package.json`](https://github.com/mapbox/assembly/blob/mb-pages/package.json).
+- Publish the new version on npm via `npm publish`.
+- Run `npm run deploy` to upload the new version to s3. **Note** you will need
+to be authenticated on AWS to do so.
+
+---
 
 For other scripts, look in `package.json`.
--- a/scripts/deploy.js
+++ b/scripts/deploy.js
@@ -1,0 +1,36 @@
+'use strict';
+
+const fs = require('fs');
+const path = require('path');
+const globby = require('globby');
+const S3 = require('aws-sdk/clients/s3');
+const pkg = require('../package');
+
+const bucket = new S3({
+  params: {
+    Bucket: 'www.mapbox.com/assembly/' + pkg.version
+  }
+});
+
+const glob = [
+  path.join(__dirname, '../dist/opensans-*'),
+  path.join(__dirname, '../dist/assembly*')
+];
+
+globby(glob)
+  .then((files) => {
+    const uploadFiles = files.map((file) => {
+      return bucket.upload({
+        Key: path.basename(file),
+        Body: fs.createReadStream(file)
+      }).promise();
+    });
+
+    return Promise.all(uploadFiles);
+  })
+  .then(() => {
+    console.log('DEPLOYED Assembly', pkg.version);
+  })
+  .catch((err) => {
+    console.log(err.stack);
+  });
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "A CSS framework",
   "main": "index.js",
   "scripts": {
+    "deploy": "npm run build && node scripts/deploy",
     "copy-site-assets": "node scripts/copy-site-assets",
     "clean-svgs": "node scripts/clean-svgs",
     "check-size": "node scripts/check-size",
@@ -54,6 +55,7 @@
     "svgstore": "^2.0.2"
   },
   "devDependencies": {
+    "aws-sdk": "^2.7.20",
     "babel-cli": "^6.18.0",
     "babel-preset-es2015": "^6.18.0",
     "babel-preset-react": "^6.16.0",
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -131,5 +131,10 @@ npm start # Builds everything, starts a dev server, rebuilds & reloads on change
 
 npm run build:svg # Build SVGs
 ```
+### Deploying
+
+- Update the version key in `package.json`.
+- `mbx auth {AUTH_CODE}`
+- `npm run deploy`
```
davidtheclark, line 144: How about changing this section title to "Releasing" and including anything related to releases in general, not just the deployment of the built code. E.g. changelog needs to be updated — anything else?

README.md
```diff
@@ -131,5 +131,10 @@ npm start # Builds everything, starts a dev server, rebuilds & reloads on change
 
 npm run build:svg # Build SVGs
 ```
+### Deploying
+
+- Update the version key in `package.json`.
+- `mbx auth {AUTH_CODE}`
```
davidtheclark: Not sure this should be in a public repo, eh?

scripts/deploy.js
```diff
@@ -0,0 +1,39 @@
...
+function upload(params) {
+  return new Promise((resolve, reject) => {
+    bucket.upload(params, (err) => {
+      if (err) reject(err);
+      return resolve();
+    });
+  });
+}
+
+execFile('find', ['./dist'], (err, files) => {
```
davidtheclark: You can use the globby library (already included) to target the files you want without later filtering them.

scripts/deploy.js
```diff
@@ -0,0 +1,39 @@
...
+
+const bucket = new S3({
+  params: {
+    Bucket: 'www.mapbox.com/assembly/' + pkg.version
+  }
+});
+
+function upload(params) {
+  return new Promise((resolve, reject) => {
+    bucket.upload(params, (err) => {
```
davidtheclark: The return value of `s3.upload` has [a `promise` property](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3/ManagedUpload.html#promise-property) that makes obviates the need for this wrapper function.

scripts/deploy.js
```diff
@@ -0,0 +1,39 @@
...
+      file.indexOf('opensans-') !== -1;
+  }).map((file) => {
+    const params = {
+      Key: file.replace('./dist/', ''),
+      Body: fs.createReadStream(file)
+    };
+    return upload(params);
+  })).then(() => {
+    console.log('DEPLOYED Assembly', pkg.version);
+  });
```
davidtheclark, line 36: We need to catch and log any errors that occur.

README.md
tristen: 👍 @davidtheclark are we going to `git tag -a v1 -m 'v1'` releases as well? We should right?

scripts/deploy.js
tristen: Ahh thanks for the link to that.

README.md
```diff
@@ -131,5 +131,15 @@ npm start # Builds everything, starts a dev server, rebuilds & reloads on change
 
 npm run build:svg # Build SVGs
 ```
+### Releasing
+
+- Document changes in the [`CHANGELOG`](https://github.com/mapbox/assembly/blob/mb-pages/CHANGELOG.md).
+- Tag the version in git and `git push --tags`
```
davidtheclark: Missing period.

README.md
```diff
@@ -131,5 +131,15 @@ npm start # Builds everything, starts a dev server, rebuilds & reloads on change
...
+- Increment the version key in [`package.json`](https://github.com/mapbox/assembly/blob/mb-pages/package.json).
+- Publish the new version on npm via `npm publish`
```
davidtheclark: Missing period.

scripts/deploy.js
```diff
@@ -0,0 +1,37 @@
...
+      return err;
+    });
+  }))
+  .catch((err) => {
+    throw err;
+  })
+  .then(() => {
+    console.log('DEPLOYED Assembly', pkg.version);
+  });
+});
```
davidtheclark: This is on the right track! But I think it can be simplified/clarified some. Here's what I'd recommend trying:

```js
// Start the Promise chain.
// Any thrown errors or (returned) rejected Promises
// in this chain will be passed all the way through
// to the catch statement.
globby(globy)
  .then((files) => {
    const uploadFiles = files.map((file) => {
      return bucket.upload({
        // There's an fs method for this
        Key: fs.basename(file),
        Body: fs.createReadStream(file)
      }).promise();
    });
    // By returning this Promise, we insert it into the
    // Promise chain, which ensures that any rejections
    // will be passed along to the next catch statement
    return Promise.all(uploadFiles);
  })
  .then(() => {
    // If things went fine, we end up here
    console.log('DEPLOYED Assembly', pkg.version);
  })
  .catch((err) => {
    // If any Promises in the chain were rejected, we'll end up here.
    // Throwing an error here does no good, because
    // we're in a Promsie chain: it just gets passed along.
    // So we log it here.
    console.log(err.stack);
  });
```

Issue #533, opened 2017-01-04
In pill example in Firefox and Safari, button text is left aligned


davidtheclark:
![screen shot 2017-01-04 at 9 49 32 am](https://cloud.githubusercontent.com/assets/628431/21650717/236a9e2e-d263-11e6-8d33-866470798d1b.png)

Issue #547, opened 2017-01-04
Buttons should not be flex


davidtheclark:
I guess that way to center things will not work:

> Certain HTML elements, like <fieldset> and <button>, do not work as flex containers. The browsers default rendering of those element's UI conflicts with the display: flex declaration.

From https://github.com/philipwalton/flexbugs#9-some-html-elements-cant-be-flex-containers

@dasulit Any chance this corresponds with the weirdness you were experiencing?

PR #555, opened 2017-01-04
Remove flex classes from buttons and recreate pills

dasulit:
@davidtheclark for review..

I'm concerned about two things:
1. that constructing vertical pills in an inline-block container will cause problems
2. that using margin-left to close the space between inline horizontal pills is too hacky

fixes https://github.com/mapbox/assembly/issues/533 and https://github.com/mapbox/assembly/issues/547

### DIFF
--- a/site/catalog/pills.js
+++ b/site/catalog/pills.js
@@ -21,21 +21,17 @@ class Pills extends React.Component {
           <button className='btn btn--red btn--stroke btn--pill btn--pill-hr round'>Confirm</button>
         </div>
 
-        <div className='mb18'>
-          <div className='flex-parent-inline flex-parent--column'>
-            <button className='btn btn--orange btn--s btn--pill btn--pill-vt'>Confirm</button>
-            <button className='btn btn--orange btn--s btn--pill btn--pill-vc'>Confirm</button>
-            <button className='btn btn--orange btn--s btn--pill btn--pill-vb is-active'>Confirm</button>
-          </div>
+        <div className='mb18 flex-parent-inline flex-parent--column'>
+          <button className='btn btn--orange btn--s btn--pill btn--pill-vt'>Confirm</button>
+          <button className='btn btn--orange btn--s btn--pill btn--pill-vc'>Confirm</button>
+          <button className='btn btn--orange btn--s btn--pill btn--pill-vb is-active'>Confirm</button>
         </div>
 
-        <div className='mb18'>
-          <div className='flex-parent-inline flex-parent--column'>
-            <button className='btn btn--stroke btn--s btn--pill btn--pill-vt round'>Confirm</button>
-            <button className='btn btn--stroke btn--s btn--pill btn--pill-vc round'>Confirm</button>
-            <button className='btn btn--stroke btn--s btn--pill btn--pill-vc round'>Confirm</button>
-            <button className='btn btn--stroke btn--s btn--pill btn--pill-vb round is-active'>Confirm</button>
-          </div>
+        <div className='mb18 ml18 flex-parent-inline flex-parent--column'>
+          <button className='btn btn--stroke btn--s btn--pill btn--pill-vt round'>Confirm</button>
+          <button className='btn btn--stroke btn--s btn--pill btn--pill-vc round'>Confirm</button>
+          <button className='btn btn--stroke btn--s btn--pill btn--pill-vc round'>Confirm</button>
+          <button className='btn btn--stroke btn--s btn--pill btn--pill-vb round is-active'>Confirm</button>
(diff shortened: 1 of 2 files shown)


site/catalog/pills.js
```diff
@@ -8,34 +8,30 @@ class Pills extends React.Component {
           Button pills
         </h2>
 
-        <div className='flex-parent mb18'>
-          <button className='btn btn--pill btn--pill-hl'>Confirm</button>
-          <button className='btn btn--pill btn--pill-hc is-active'>Confirm</button>
-          <button className='btn btn--pill btn--pill-hr'>Confirm</button>
+        <div className='mb18'>
+          <button className="btn btn--pill btn--pill-hl">Confirm</button>
```
davidtheclark: Throughout the examples we use single quotation marks in the HTML. Not very significant, but may as well change it back, eh?

src/buttons.css
```diff
@@ -147,18 +144,15 @@
...
  * @example
- * <div>
- *   <div class="mr12 flex-parent-inline">
- *     <button class="btn btn--pill btn--pill-hl">Left</button>
- *     <button class="btn btn--pill btn--pill-hc">Center</button>
- *     <button class="btn btn--pill btn--pill-hr">Right</button>
- *   </div>
+ * <button class="btn btn--pill btn--pill-hl">Left</button>
+ * <button class="btn btn--pill btn--pill-hc">Center</button>
+ * <button class="btn btn--pill btn--pill-hr">Right</button>
```
davidtheclark: If I try adding `btn--stroke` to this class list we get this space between the buttons:

screen shot 2017-01-04 at 5 59 05 pm

src/buttons.css
```diff
@@ -26,6 +26,7 @@
  * </div>
  */
 .btn {
+  display: inline;
```
davidtheclark: I think the browser default style is `inline-block`, so if you're removing the `display: inline-flex` declaration, I don't think you have to replace it with `display: inline`.

src/buttons.css
```diff
@@ -174,12 +168,14 @@ active color is visible on all borders */
 /* Extra .btn in selectors below to increase specificity
 against .round class */
 
-/* 1px margins create gutters. -2px margins blend borders */
+/* -3px margins on horizontal pills work around space created
+between inline elements, while maintaining gutters */
```
davidtheclark: I think this is a real problem ... the space can vary based on browser, font, and maybe some other factors. This was one reason to use a flex parent. Why do you think we should remove the flex parent?

src/buttons.css
```diff
@@ -190,31 +186,37 @@ against .round class */
...
+  display: block;
+  width: 100%;
 }
 
 .btn.btn--pill-vb {
   border-top-right-radius: 0 !important;
   border-top-left-radius: 0 !important;
+  display: block;
+  width: 100%;
 }
```
davidtheclark, line 211: Here we could safely use a fancy attribute selector, I think, if we really wanted to:

```css
.btn[class*='btn--pill-'] {
  display: block;
  width: 100%;
}
```

davidtheclark (commented):
Thanks for taking this on!

Your removal of Flex from the buttons seems good. My main question about the pill changes: Why did you feel the need to remove the flex parent? What trouble is it causing?

dasulit:
@davidtheclark ah, I got carried away trying to recreate these entirely without flex. I added the flex containers back, is this all right?

src/buttons.css
```diff
@@ -178,8 +174,8 @@ against .round class */
 
 .btn.btn--pill-hc {
   border-radius: 0 !important;
-  margin-left: 1px;
-  margin-right: 1px;
+  margin-left: 1px !important;
+  margin-right: 1px !important;
```
davidtheclark: I believe that all the `!important`s you've added to margin values are unnecessary.

src/buttons.css
```diff
@@ -194,34 +190,40 @@ against .round class */
...
+  margin-bottom: 1px !important;
+  display: block !important;
+  width: 100% !important;
 }
 
 .btn.btn--pill-vt {
   border-bottom-right-radius: 0 !important;
   border-bottom-left-radius: 0 !important;
+  display: block !important;
+  width: 100% !important;
```
davidtheclark: I think that these `display` and `width` properties do not need to be `!important`.

I believe our convention has been that we use `!important` when a declaration corresponds directly with the class name, and otherwise we avoid `!important` unless it's necessary. It is necessary here for the `border-radius` declarations, because we need them to be able to override `round-*` utility classes. But I don't think we need `!important` for other parts.

Issue #541, opened 2017-01-04
On mobile Safari, empty search input has full border curve and inner box shadow


davidtheclark:
![image](https://cloud.githubusercontent.com/assets/628431/21651710/e45387ec-d266-11e6-8be1-c80ac973fc8e.png)

The border curve goes away when you focus the input, which is extra weird. But the inner box shadow remains:

![image](https://cloud.githubusercontent.com/assets/628431/21651722/fa7aca94-d266-11e6-81f0-2836759d3040.png)

Issue #540, opened 2017-01-04
On mobile Safari, date input is not same width as other inputs


davidtheclark:
![image](https://cloud.githubusercontent.com/assets/628431/21651667/be0d5612-d266-11e6-99da-fbf8ee3a6181.png)

davidtheclark:
As far as I can tell, the only way to fix this would be to set `display: block; width: 100%` on `.input` elements. In which case we'd want to do the same thing on `.textarea` elements.

This is a significant change, but I'm not opposed to it — kind of leaned that way earlier on but didn't say anything. The justification would be that we are forcing the user to deliberately decide on an input width, either by assigning a width class or wrapping the input in a container that determines its width. The browser default widths are definitely inconsistent in this case, maybe also in other cases.

Alternately, we could just warn people not to use `type="date"` inputs, and maybe prevent any styling from being applied to `type="date"` inputs.

Thoughts?

samanpwbb:
@davidtheclark i agree with you about 100%'ing inputs and textareas.

PR #558, opened 2017-01-05
Fix appearance of search inputs on mobile Safari

dasulit:
adding `-webkit-appearance: none` to search inputs fixes https://github.com/mapbox/assembly/issues/541

I'm continuing to investigate other Safari bugs, but we can merge for beta if we want.

### DIFF
--- a/src/forms.css
+++ b/src/forms.css
@@ -58,7 +58,7 @@
   height: 0;
 }
 .input[type='search'] {
-  appearance: textfield;
+  appearance: none;
 }
 .input[type='search']::-webkit-search-decoration,
 .input[type='search']::-webkit-search-cancel-button {


src/forms.css
```diff
@@ -59,6 +59,7 @@
 }
 .input[type='search'] {
   appearance: textfield;
+  -webkit-appearance: none;
```
davidtheclark: `appearance` and `-webkit-appearance` clash here: `-webkit-appearance` is populated by Autoprefixer when we use `appearance`. So if you think `none` is a better value, we should just use `appearance: none`.

dasulit: This was a facepalm moment, I fixed! I would expect that this would fix the other Safari input bug https://github.com/mapbox/assembly/issues/540 but it did not 👎

davidtheclark:
I incorporated this fix into #577.

PR #560, opened 2017-01-05
Adds flex-child class. Removes uses of flex shorthand.

### LINKED ISSUE #548: Maybe add a flex-child class
After realizing #547, I thought I'd run through all the [documented flexbugs](https://github.com/philipwalton/flexbugs) and see if any others might bite users of our utility classes.

I think we're in pretty good shape, given the limited scope of our flexbox classes. I think two changes might help us avoid these bugs (pretty much *in IE11 only*):

- Add a `flex-child` class. It *kind* of sucks to do this, but it wouldn't be too bad. Usually things will work fine *without* applying this class to children; but it might save us sometimes. This would also involve changing the existing `flex-child-*` class to be modifiers: `flex-child--*`. This class would have just a few bug-avoiding declarations:
  - `flex-shrink: 0;` (re https://github.com/philipwalton/flexbugs#1-minimum-content-sizing-of-flex-items-not-honored)
  - `flex-basis: auto;` (re https://github.com/philipwalton/flexbugs#1-minimum-content-sizing-of-flex-items-not-honored)
  - `max-width: 100%;` (re https://github.com/philipwalton/flexbugs#2-column-flex-items-set-to-align-itemscenter-overflow-their-container)
  - `display: block;` (re https://github.com/philipwalton/flexbugs#12-inline-elements-are-not-treated-as-flex-items)
- Remove any usage of the `flex` shorthand (re https://github.com/philipwalton/flexbugs#6-the-default-flex-value-has-changed)

Anybody have any concerns about these changes?

davidtheclark:
Changes existing `flex-child-` classes to modifiers, with the prefix `flex-child--`. Removes `flex-child-noshrink` because `flex-child` covers that property already. All this to anticipate flexbugs.

Closes #548.

I also removed a few `!important` that I think were unnecessary, because they did not correspond directly to the class name.

@samanpwbb for review.

### DIFF
--- a/site/navigation.js
+++ b/site/navigation.js
@@ -61,10 +61,10 @@ class Navigation extends React.Component {
         <a href='/assembly/' className='mt24 mb18 mx24 hmin48 link link--purple block'>
          <Logo className='icon icon--l' />
         </a>
-        <div className='flex-child-grow-mm scroll-auto pr18 pl24'>
+        <div className='flex-child flex-child--grow-mm scroll-auto pr18 pl24'>
           {navEls}
         </div>
-        <div className='mt12 mb24 mx24 flex-child-noshrink-mm'>
+        <div className='mt12 mb24 mx24'>
           <a className='color-purple-on-hover txt-s' href='https://github.com/mapbox/assembly/'>
             <svg
               className='icon'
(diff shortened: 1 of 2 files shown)


samanpwbb (approved):
All these changes look good

PR #561, opened 2017-01-05
rounded table

### LINKED ISSUE #559: add .round table class
screen shot 2017-01-05 at 9 49 17 am

mayagao:
closes #559

screen shot 2017-01-05 at 5 12 35 pm

@davidtheclark review

### DIFF
--- a/src/tables.css
+++ b/src/tables.css
@@ -36,10 +36,15 @@
 .prose table {
   width: 100%;
   background-color: transparent;
-  border-collapse: collapse;
   border-spacing: 0;
+  border-collapse: separate;
   border: 1px solid var(--gray-light);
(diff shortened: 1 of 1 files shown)


davidtheclark (changes requested):
Huh, seems like we need clever trickery to be both rounded borders and collapsed borders. Nice job figuring that out.

However, I think something must be not quite right about the selectors yet, because I'm seeing different appearances between `.table` and `.prose table`. From the debug page:

screen shot 2017-01-05 at 9 07 37 am

davidtheclark:
@mayagao I am going to try finishing this up, since you are sleeping.

davidtheclark (approved):

PR #572, opened 2017-01-05
New IE friendly spinners

dasulit:
A square spinner, which is a pretty big departure from spinner 1.0. It also works fine without a `--dark` modifier.

@samanpwbb does this look cool enough for Assembly? 😕

### DIFF
--- a/src/miscellaneous.css
+++ b/src/miscellaneous.css
@@ -14,32 +14,39 @@
  * @example
  * <div class="loading"></div>
  * <div class="loading loading--s"></div>
- * <div class="loading loading--dark bg-gray-dark"></div>
+ * <div class="bg-gray-dark">
+ *   <div class="loading loading--dark"></div>
+ *   <div class="loading loading--dark loading--s"></div>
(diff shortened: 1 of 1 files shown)


samanpwbb:
I liked the old spinners! I find this one a little bit disorienting.

I haven't looked to see how bad IE actually is yet, doing so now.

samanpwbb:
Hm, the shake is pretty prominent. Can we try using SVG?

dasulit:
I'm on it! This was worth a try, but might as well not introduce a new design at this point.

dasulit:
Dang it, I updated to SVGs, but they don't show up in IE11 at all.

davidtheclark:
> they don't show up in IE11 at all.

I'm pretty sure this wasn't the case last time I looked in IE11, so something may have changed. I'll take another look ...

dasulit:
@davidtheclark may you review this when you get a chance?

davidtheclark:
Looks great! I'll merge.

PR #578, opened 2017-01-06
Add negative margin classes

### LINKED ISSUE #518: Add negative margin rules
Negative margins are generally useful, especially for absolute positioning and then offsetting things inside a relative-positioned container. 

I think I would use rules like `m-neg12`, `m-neg18`, and `m-neg24` pretty regularly if we had them.

davidtheclark:
Closes #518 by adding negative margin classes for 1, 2, 3, 6, 12, 18, 24.

1, 2, 3 are there, even though they don't fit the standard scale, because they will probably be some of the most frequently used, to overlap borders or small padding.

This change increases gzipped CSS size from 23.9 kB to 24.4 kB, ~2%. @samanpwbb what do you think — worth it?

### DIFF
--- a/src/layout.css
+++ b/src/layout.css
@@ -786,6 +786,13 @@
  * <div class='mt24 bg-darken10'>mt24</div>
  * @memberof Margins
  */
+.mt-neg1 { margin-top: -1px !important; }
+.mt-neg2 { margin-top: -2px !important; }
+.mt-neg3 { margin-top: -3px !important; }
+.mt-neg6 { margin-top: -6px !important; }
+.mt-neg12 { margin-top: -12px !important; }
+.mt-neg18 { margin-top: -18px !important; }
+.mt-neg24 { margin-top: -24px !important; }
 .mt0 { margin-top: 0 !important; }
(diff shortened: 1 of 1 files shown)


davidtheclark:
One thought I had, if the increase makes anybody squeamish, is that I wonder how much negative margins need to be responsive. If they don't need to be responsive, we could just people to use inline styles — basically the same thing without the responsive.

samanpwbb:
This is another one of those weird decisions. There's as much reason to make these responsive as anything else. There's also reason to include even larger negative margins. Kind of? Or can we suggest https://github.com/mapbox/assembly/blob/a24e70b191cb90b315faa99645b83e3b70718712/src/layout.css#L2205-L2216 when you want margins that actually effect layout?

We could say, use bleed classes when you really want to pull something out of it's parent, but for repositioning icons and doing other minor layout tweaks, use these negative margin classes?

davidtheclark:
@samanpwbb That's what I was thinking. I don't recall using largish negative margins myself much except for the full-screen things bleed is made for. If you've had other experiences, though, they should be honored.

Issue #581, opened 2017-01-07
Range inputs are not vertically aligned


davidtheclark:
![screen shot 2017-01-06 at 5 10 03 pm](https://cloud.githubusercontent.com/assets/628431/21737097/0f74ea4c-d433-11e6-8e1e-0e1bc5aa959e.png)

We should try to make range inputs occupy the same vertical space as text inputs, in large and small sizes. We may be able to do this by adding a negative margin to the top.

PR #584, opened 2017-01-07
Fix range input IE-focused bugs

### LINKED ISSUE #524: Range dot is off-center in Edge
screen shot 2017-01-04 at 8 41 52 am

### LINKED ISSUE #582: Range input thumbs not colored in IE11
screen shot 2017-01-06 at 5 14 46 pm

They *do* look colorful in Edge.

### LINKED ISSUE #583: Range track has sharp edges in IE11 and Edge
screen shot 2017-01-06 at 5 18 49 pm

davidtheclark:
- Closes #582 by fixing a typo.
- Closes #524 by magic.
- Closes #583 by addressing a little oversight.
- Gives up on #581 so "closes" it by adding a note to the documentation.

@samanpwbb for review.

### DIFF
--- a/scripts/build-color-variants.js
+++ b/scripts/build-color-variants.js
@@ -339,23 +339,23 @@ function buildColorVariants(variables, config) {
       // Set the track and thumb color.
       // Set the track and thumb color to dark on hover.
       return result += stripIndent(`
-        .range--${color}::-webkit-slider-runnable-track { background: ${colorValue}; }
-        .range--${color}::-moz-range-track { background: ${colorValue}; }
-        .range--${color}::-ms-fill-lower { background: ${colorValue}; }
-        .range--${color}::-ms-fill-upper { background: ${colorValue}; }
+        .range--${color} > input::-webkit-slider-runnable-track { background: ${colorValue}; }
+        .range--${color} > input::-moz-range-track { background: ${colorValue}; }
+        .range--${color} > input::-ms-fill-lower { background: ${colorValue}; }
+        .range--${color} > input::-ms-fill-upper { background: ${colorValue}; }
 
-        .range--${color}::-webkit-slider-thumb { border-color: ${colorValue}; }
-        .range--${color}::-ms-thumb { border-color: ${colorValue}; }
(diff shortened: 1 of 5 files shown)


samanpwbb:
@davidtheclark could we center range inputs by adding:

```css
.range {
  height: 24px;
  vertical-align: middle;
}

.range--s { 
  height: 18px;
}
```

davidtheclark:
@samanpwbb `vertical-align: middle` only helps if the element it's aligning with also has `vertical-align: middle`. So I'm not sure that's a more robust solution than just recommending vertical centering patterns (whether that's flexbox, or `inline-block align-middle` on the whole row of things).

samanpwbb:
@davidtheclark If you apply `vertical-align: middle` to range inputs and place them inline or inline block alongside text or other inputs, they correctly center themselves – no need for setting `vertical-align: center` on other elements.

samanpwbb:
I looked into centering a little more and I'm having a hard time figuring out why this happens (regardless of whether anything is `inline` or `inline-block`):
screen shot 2017-01-10 at 10 03 56 pm

Also, it seems like the `vertical-align: middle;` rule on the track (in `-moz` and  `-webkit` at least) doesn't do anything.

samanpwbb (approved):
I think we can merge this as-is and lean on flex alignment... without it, range input baseline alignment is pretty whacky and so I'm a bit dissatisfied with this. I don't see any solutions after experimenting a little. Seems like range inputs do their own thing a little bit WRT display properties and line height.

samanpwbb:
Okay. I give up!

I do have a few pieces of feedback:

- moz range track  is 3px, others are 4px.
- moz thumbs (didn't check ie/edge) should be sized the same as webkit handles (20px/20px instead of 17px/17px), but they need `box-sizing: border-box` to be declared in order for that to work.
- inline / inline block inputs in IE end up positioned higher than in other browsers. Maybe we can live with this. 
- `vertical-align: middle;` doesn't appear to do anything on tracks.

davidtheclark:
I have added a range-wrapping div. This ensures that *across browsers* ranges are the same height as other inputs and are vertically centered. I think this is the way to go.

PR #586, opened 2017-01-09
Add infinite modifier for animations

### LINKED ISSUE #536: fade-in, fade-out, and shake examples are useless
By the time you get to these, they have already run their course and illustrate nothing. Maybe the best we can do is add a note that this is the case and encourage the user to refresh their browser at that scroll position?

tristen:
- Closes #536

### DIFF
--- a/src/animations.css
+++ b/src/animations.css
@@ -5,51 +5,53 @@
  */
 
 /**
- * Infinite animations.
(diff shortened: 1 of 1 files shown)


samanpwbb (approved):
Nice, looks like you caught a bug in here too 👍

PR #597, opened 2017-01-10
Capture :hover too when .is-active is used

### LINKED ISSUE #596: .{property}-{color}-on-active.is-active should override hover rules
That's a crazy long classname but for the hover state of active elements that have a different style than it's siblings:

on-hover-is-active

tristen:
- Fixes #596

### DIFF
--- a/scripts/build-color-variants.js
+++ b/scripts/build-color-variants.js
@@ -480,11 +480,13 @@ function buildColorVariants(variables, config) {
       const colorValue = variables[color];
       return result += stripIndent(`
         .shadow-${color}-on-hover:hover,
-        .shadow-${color}-on-active.is-active {
+        .shadow-${color}-on-active.is-active,
+        .shadow-${color}-on-active.is-active:hover {
           box-shadow: 0 0 10px 2px ${colorValue} !important;
         }
         .shadow-bold-${color}-on-hover:hover,
-        .shadow-bold-${color}-on-active.is-active {
+        .shadow-bold-${color}-on-active.is-active,
+        .shadow-bold-${color}-on-active.is-active:hover {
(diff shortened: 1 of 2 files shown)


scripts/build-color-variants.js
```diff
@@ -480,11 +480,13 @@ function buildColorVariants(variables, config) {
       const colorValue = variables[color];
       return result += stripIndent(`
         .shadow-${color}-on-hover:hover,
-        .shadow-${color}-on-active.is-active {
+        .shadow-${color}-on-active.is-active,
+        .shadow-${color}-on-active.is-active:hover {
```
davidtheclark, line 484: Is it not possible to accomplish the same thing by re-ordering? That could be more efficient.

tristen: How would reordering help? `:hover` still wouldn't be captured.

davidtheclark: `.x:hover` and `.x.y` have the same specificity. So whichever comes later in the stylesheet should override. You then don't need to create rules for `.x.y:hover`.

tristen: 😭 I thought the specificity was slightly higher when targeting a pseudo selector. Sorry about that. will fix. 😭

tristen: Hmm looks like they need to be decoupled from [being grouped together](https://github.com/mapbox/assembly/blob/573999d2bbdfd4f5ce9d4d2d95b9199cb96dfcd4/test/__snapshots__/build-color-variants.jest.js.snap#L9-L12) which means more lines.

tristen: Discussed in chat with @davidtheclark: this looks like the better path to go down.

PR #600, opened 2017-01-10
Add hide-visually

### LINKED ISSUE #594: Reconsider method for hiding input elements of radios and checkboxes
I just read [an article](https://gdstechnology.blog.gov.uk/2016/12/13/improving-forms-with-new-radios-and-checkboxes/) about styling radios/checkboxes essentially the same way we have. But they note one problem they ran into that we could address: 

> Version 12 worked fine with the new radio buttons, but version 13 failed to recognise the inputs – the user couldn’t click on them. We narrowed this down to the code we were using to hide the native controls. As soon as we reduced the control in size to near-invisibility, or positioned the control outside off-screen NaturallySpeaking was unable to interact with it. We solved this issue by making the control completely transparent, which meant NaturallySpeaking 13 was able to recognise it.

By inspecting [their example page](https://govuk-prototype-kit.herokuapp.com/docs/examples/elements/radio-buttons-checkboxes) we can see that the actual `<input>` element is not placed offscreen, but is instead overlapping with the display element but transparent. (We do the same things with `<select>` elements right now.)

davidtheclark:
Use the same technique to hide toggly input elements.

Closes #594.

I ended up deciding to use the same technique [as in HTML5 boilerplate](https://github.com/h5bp/html5-boilerplate/blob/master/dist/css/main.css#L135), since that's at least semi-authoritative, probably well tested.

I *could* reduce a little CSS by making people add the `hide-visually` class to their `<input>` elements for these components. But that would be a bigger change, more of a hassle generally, and this is some very short repetition.

@samanpwbb for review.

### DIFF
--- a/src/forms.css
+++ b/src/forms.css
@@ -517,12 +517,20 @@ because it seems to be more cross-browser consistent with all HTML5 input types 
   display: inline-flex;
 }
 
+/* Same as the hide-visually class */
 .toggle-container > input,
 .checkbox-container > input,
 .switch-container > input,
 .radio-container > input {
+  border: 0;
+  clip: rect(0 0 0 0);
+  height: 1px;
+  margin: -1px;
+  overflow: hidden;
+  padding: 0;
   position: absolute;
-  left: -9999px;
+  width: 1px;
+  white-space: nowrap;
 }
 
 .checkbox,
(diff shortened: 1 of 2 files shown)


samanpwbb (approved):
Great

PR #601, opened 2017-01-10
Add Layout Scales page

### LINKED ISSUE #551: Scale table
In chat @tmcw presented a table showing the size variants for our different "sized" things.

I think this would be a generally useful reference — a good page to add to the documentation site. When I'm writing a class and thinking "What widths are available?", I would much rather go to one compact page and scan a table than look seek out and expand a class list on the huge documentation page.

davidtheclark:
Closes #551.

Creates a Layout Scales page. It's not too pretty, but it's legible. What do you think?

@dasulit for review. (Better design ideas?)

screenshot-localhost 3000-2017-01-10-16-59-48


dasulit:
I already know I'm going to be using this page a lot!

As far as design ideas... since you asked, I have a couple, take 'em or leave 'em. I'm mainly thinking about how we can make this easier to quickly scan.

I suggest that we make the variable content extend in the vertical direction. This would get rid of unpredictable line breaks in the lists of scales. Something like this could reflect the initial effectiveness of the table Tom shared in chat:

screen shot 2017-01-10 at 8 27 04

While this layout ^ would be my recommendation, if reorganizing would be too much of a lift, a quicker suggestion would be to introduce a table layout to what you have currently:

screen shot 2017-01-10 at 8 38 38

site/layout_scales.js
```diff
@@ -0,0 +1,123 @@
...
+    return match.sub[0];
+  });
+}
+
+const classSets = {
+  'Grid Gutters': {
+    classPatterns: ['grid--gut{n}'],
+    numbers: findNumbers(/\.grid--gut(\d+) {/g)
+  },
+  'Z-Indexes': {
```
dasulit, line 20: I'm feeling like we could remove z-indexes from this page— the pattern is much more memorable than the other scales being documented on this page, and z-indexes are the only non-pixel unit?

davidtheclark: Yeah, I guess you're right.

davidtheclark:
@dasulit The reason I veered away from the vertical stacks is that it becomes awkward when you have multiple rows of sections — a lot of empty white-space can appear, right? Imagining how your partial mock would look with multiple rows, because 10 items — and responsive — do you still think it's the better option?

dasulit:
Ah, I'm having a hard time imagining what you mean by "empty white-space"— are you referring to the space created by columns of different heights? Demonstrated with a blue rectangle here:

screen shot 2017-01-10 at 8 27 04

If that's what you mean, that gap doesn't bother me too much. On a narrower screen the columns could be full-width, which would result in a longer page, but I think that would be acceptable.

davidtheclark:
Yeah, that is what I meant by "empty white-space". I can try this out tomorrow and post another screenshot.

davidtheclark:
@dasulit Here's an approximation of your suggestion. I'm not sure it's worth the trouble to separate out negative classes; just left those in the same column. What do you think?

screenshot-localhost 3000-2017-01-11-09-12-43

samanpwbb:
I preferred the horizontally aligned design – some horizontal scroll might make sense here to keep everything on one line.

davidtheclark:
@samanpwbb Do you mean a column for each item but a horizontal scroll across them? Or the original design without columns?

samanpwbb:
@davidtheclark I think the original design works better. My suggestion is to do this:

screen shot 2017-01-11 at 11 18 05 am

davidtheclark:
I'm ending up with this:

screen shot 2017-01-11 at 9 38 13 am

I agree with @dasulit that the borders around numbers help for scanning. A horizontal table won't really work, though, because of wrapping, so I set each cell off a little.

Anybody who wants to revisit the design is welcome to after this is merged!

PR #607, opened 2017-01-11
remove hover state from forms, closes #606

### LINKED ISSUE #606: Text inputs and textareas should have different focus state vs hover state
I think text inputs and textareas are one place where a focus state always makes sense. It is _such_ a convention that it feels wrong to not have this.

We could:

- Add 2-4 px inner shadow on text inputs by default.
- Add some `on-focus` modifiers just for text inputs and textareas.

samanpwbb:
Nowhere else are there hover states on text inputs / textareas. This is what was bothering me when I opened https://github.com/mapbox/assembly/issues/606.

### DIFF
--- a/src/forms.css
+++ b/src/forms.css
@@ -37,8 +37,6 @@
   width: 100%;
 }
 
-.input:hover,
-.textarea:hover,
 .input:focus,
 .textarea:focus {
   border-color: var(--default-primary-interactive-color);


samanpwbb:
@davidtheclark for review.

davidtheclark (approved):

davidtheclark:
🆗

PR #611, opened 2017-01-11
Include a height flex-parent

tristen:
To demonstrate how flex-child--grow works.


davidtheclark:
@tristen What does the height add that wasn't already there? `flex-child--grow` should make the child fill the horizontal space — which is what it's doing, right? Just curious what's missing.

tristen:
ok

PR #617, opened 2017-01-12
300 360 420

### LINKED ISSUE #613: 60px increments from 120 to 480 for margin, width, and height utility classes
That would introduce `300`,  `360`, & `420` to complete above. Currently the jump from `240` to `480` is high and both sizes don't work great for fixed width sidebars in single page apps. `240` is too narrow and `480` is too wide for most laptop screens.

tristen:
🍁 🍁 🍁 🍁 🍁 🍁 

Closes #613

### DIFF
--- a/src/layout.css
+++ b/src/layout.css
@@ -571,6 +571,9 @@
 .m120 { margin: 120px !important; }
 .m180 { margin: 180px !important; }
 .m240 { margin: 240px !important; }
+.m300 { margin: 300px !important; }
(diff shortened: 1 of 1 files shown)


samanpwbb:
It would be more consistent to add `30` and `42` to this PR as well, so the small sizes and large sizes parellel each other.

tristen:
Closing in favour of https://github.com/mapbox/assembly/pull/618

PR #618, opened 2017-01-13
Generate scales

### LINKED ISSUE #613: 60px increments from 120 to 480 for margin, width, and height utility classes

tristen:
Generates scales programatically.

Closes #613


scripts/build-layout-scales.js
```diff
@@ -0,0 +1,114 @@
+'use strict';
+
+const stripIndent = require('strip-indent');
+const mediaQueries = require('../src/media-queries');
+const allScales = require('../src/scales');
+
+function buildLayoutScales(config) {
+  const layoutScales = typeof config === 'object' ? config : Object.assign(allScales, config);
```
davidtheclark: I'm confused about this line. What's the aim?

scripts/build-layout-scales.js
```diff
@@ -0,0 +1,114 @@
...
+    }, '');
+    css += '/** @endgroup */\n';
+    css += scales.reduce((result, scale) => {
+      return result += stripIndent(`
+        .grid--gut${scale} > .col {
+          padding-left: ${scale}px !important;
+        }
+      `);
+    }, '');
+    css += '\n@media (' + mediaQueries['--m-screen'] + ') {\n';
```
davidtheclark: You *could* use template literals here, too: `\n@media (${mediaQueries['--m-screen']}) {\n`.

scripts/build-layout-scales.js
```diff
@@ -0,0 +1,114 @@
...
+      `);
+    }, '');
+    css += scales.reduce((result, scale) => {
+      return result += stripIndent(`
+        .grid--gut${scale}-mm > .col {
+          padding-left: ${scale}px !important;
+        }
+      `);
+    }, '\n');
+    css += '\n}\n';
```
davidtheclark: This would probably be a little more clear & legible with an arrangement like:

```js
const medium = scales.reduce(..);
const mediumColumns = scales.reduce(..);
css += stripIndent(`
  @media (${mediaQueries['--m-screen']}) {
    ${medium}
    ${mediumColumns}
  }
`)
```

But maybe this is just quibbly.

scripts/build-layout-scales.js
```diff
@@ -0,0 +1,114 @@
...
+    result += variantGenerators[d](scales);
+  });
+
+  return result;
+}
+
+module.exports = buildLayoutScales;
+
+if (require.main === module) {
+  buildLayoutScales().catch((err) => console.error(err.stack));
```
davidtheclark: `buildLayoutScales` is not returning a promise (right?) so I think this `catch` doesn't make sense. If we have the same situation in `buildColorVariants` I think it's probably a mistake.

scripts/build-layout-scales.js
```diff
@@ -0,0 +1,114 @@
...
+    }, '\n');
+    css += '\n}\n';
+
+    return css;
+  };
+
+  let result = '\n/* Layout scales */\n';
+
+  Object.keys(variantGenerators).forEach((d) => {
+    const scales = layoutScales[d] || layoutScales.default;
```
davidtheclark: I'm not sure if we want to build in the same `default` functionality for layout scales as we have for colors. Do you think it's useful?

src/scales.json
```diff
@@ -0,0 +1,13 @@
...
+  "multiMargin": [ 0, 3, 6, 12, 18, 24, 30, 36, 42, 48, 72, 96, 120, 180, 240, 300, 360, 420, 480 ],
+  "singleMargin": [ "-neg1", "-neg2", "-neg3", "-neg6", "-neg12", "-neg18", "-neg24", 0, 3, 6, 12, 18, 24, 30, 36, 42, 48, 72, 96, 120, 180, 240, 300, 360, 420, 480 ],
+  "padding": [0, 3, 6, 12, 18, 24, 30, 36, 42, 48, 72, 96],
+  "width": [ 0, 3, 6, 12, 18, 24, 30, 36, 42, 48, 72, 96, 120, 180, 240, 300, 360, 420, 480, 720, 960, "-full" ],
+  "maxWidth": [ 0, 3, 6, 12, 18, 24, 30, 36, 42, 48, 72, 96, 120, 180, 240, 300, 360, 420, 480, 720, 960, 1200, 1500, "-full" ],
+  "minWidth": [ 0, 3, 6, 12, 18, 24, 30, 36, 42, 48, 72, 96, 120, 180, 240, 300, 360, 420, 480, "-full" ],
+  "height": [ 0, 3, 6, 12, 18, 24, 30, 36, 42, 48, 72, 96, 120, 180, 240, 300, 360, 420, 480, "-full" ],
+  "maxHeight": [ 0, 3, 6, 12, 18, 24, 30, 36, 42, 48, 72, 96, 120, 180, 240, 300, 360, 420, 480, "-full" ],
+  "minHeight": [ 0, 3, 6, 12, 18, 24, 30, 36, 42, 48, 72, 96, 120, 180, 240, 300, 360, 420, 480, "-full" ]
+}
```
davidtheclark, line 12: Oooh I like the creation of this file.

davidtheclark (commented):
@tristen I think this is on the right track! I left a couple of questions and stylistic suggestions, but really it looks to me like you've got it.

scripts/build-layout-scales.js
tristen: Ah this is a little half hearted. if a custom `config` exists, use that instead of the default `scales`. It's untested and I may have misinterpreted how you were doing this for color.

scripts/build-layout-scales.js
tristen: yeah. it's probably not. I'm happy to ditch it!

scripts/build-layout-scales.js
davidtheclark: Ok, if `config` is going to be an object or `undefined`, I think you can just do `const layoutScales = Object.assign(allScales, config)`.

scripts/build-layout-scales.js
davidtheclark: Ok, let's ditch it, we we just have `const scales = layoutScales[d]`.

scripts/build-layout-scales.js
tristen: I don't think I need to pass `config` I just omitted it.

tristen:
@davidtheclark I think this is in a good place now!

scripts/build-layout-scales.js
```diff
@@ -0,0 +1,1271 @@
...
+    css += scales.reduce((result, scale) => {
+      return result += stripIndent(`
+        .z${scale} {
+          z-index: ${value(scale)} !important;
+        }
+      `);
+    }, '');
+    css += '/** @endgroup */\n';
+
+    css += stripIndent(`\n\n@media ${mediaQueries['--m-screen']} {`);
```
davidtheclark: No need for `stripIndent` on these single-line strings that have no indent to strip.

scripts/build-layout-scales.js
```diff
@@ -0,0 +1,1271 @@
...
+
+    css += stripIndent(`\n\n@media ${mediaQueries['--xl-screen']} {`);
+    css += scales.reduce((result, scale) => {
+      return result += stripIndent(`
+        .z${scale}-mxl {
+          z-index: ${value(scale)} !important;
+        }
+      `);
+    }, '');
+    css += stripIndent('}');
```
davidtheclark: Let's stick to one pattern for inserting things into media queries — either the pattern you have above for gutters, or this one — but not both.

scripts/build-layout-scales.js
```diff
@@ -0,0 +1,1271 @@
...
+      return result += stripIndent(`
+        .my${scale} {
+          margin-top: ${value(scale)} !important;
+          margin-bottom: ${value(scale)} !important;
+        }
+      `);
+    }, '');
+    css += '/** @endgroup */\n';
+
+    css += stripIndent(`\n\n@media ${mediaQueries['--m-screen']} {`);
```
davidtheclark: This might be a totally menial optimization, but we *could*, if we were interested, group media queries pretty easily with this build script. Instead of adding rules into these individual queries you could push them into a an array (`mediumScreenRules`) and then build one media query block after you've run through all these.

scripts/build-layout-scales.js
```diff
@@ -0,0 +1,1271 @@
...
+        .my${scale}-mxl {
+          margin-top: ${value(scale)} !important;
+          margin-bottom: ${value(scale)} !important;
+        }
+      `);
+    }, '');
+    css += stripIndent('}');
+
+    css += stripIndent(`
+    \n/**
```
davidtheclark, line 134: `\n` unnecessary.

scripts/build-layout-scales.js
```diff
@@ -0,0 +1,1271 @@
...
+      `);
+    }, '');
+    css += stripIndent('}');
+
+    return css;
+  };
+
+  variantGenerators.singleMargin = function (scales) {
+    function value(v) {
+      return typeof v === 'string' ? `-${v[v.length - 1]}px` : `${v}px`;
```
davidtheclark: This seems wrong to me ... If the type is a string, e.g. `-full`, you don't want to prepend a hyphen or append a `px` unit, right?

tristen: `-full` isn't a thing for margins

scripts/build-layout-scales.js
```diff
@@ -0,0 +1,1271 @@
...
+    css += scales.reduce((result, scale) => {
+      return result += stripIndent(`
+        .w${scale}-mm {
+          width: ${value(scale)} !important;
+        }
+      `);
+    }, '');
+    css += stripIndent('}');
+
+    css += stripIndent(`\n\n@media ${mediaQueries['--l-screen']} {`);
```
davidtheclark: Ok, I'm pretty convinced by this point that there should be a helper function for generating the media query variants, instead of repeating the same pattern so many times.

scripts/build-layout-scales.js
```diff
@@ -0,0 +1,1271 @@
...
+      `);
+    }, '');
+    css += stripIndent('}');
+
+    return css;
+  };
+
+  variantGenerators.maxHeight = function (scales) {
+    function value(v) {
+      return v === '-full' ? '100%' : `${v}px`;
```
davidtheclark: There should be a shared helper that translates entries into CSS values that is used by every one of these, right? `-full` always means `100%`, `-neg{n}` always means `-{n}px`.

PR #625, opened 2017-01-15
Example docs

### LINKED ISSUE #145: Modal styles
Do we want component classes for modals?

### LINKED ISSUE #157: Badges/tokens/chips
Do we want component styles for these things?

image

### LINKED ISSUE #261: Fancy input examples
@samanpwbb suggested that he'd like to be able to do these things:

- text input with button
- text input with icon
- text input with spinner

We need to write examples showing how these are done, or if necessary a component class or two.

### LINKED ISSUE #442: Legend examples
Things like this:

screen shot 2016-12-23 at 8 06 14 am
screen shot 2016-12-23 at 8 06 56 am
screen shot 2016-12-23 at 8 07 02 am
screen shot 2016-12-23 at 9 22 16 am

It would be great to have just a standard style with variations like:

- Radius legend for point size
- Linear gradient ramp 
- Bars with values below each one

### LINKED ISSUE #587: Add a library of example layouts and components
Assembly isn't very useful until you know how to combine classes to get the design you want. We should build a `library` of markup snippets. Goals:

- Examples should be attractive and usable without any modification.
- Aim for a neutral, white-dominant style.
- Seek the simplest and most flexible markup pattern for the particular example.
- Make all examples responsive.
- Use the minimum number of classes necessary – we should expect users to _add on classes_ to our library examples in order to fit the look and feel of their site.

tristen:
Sets up pages for examples! Also adds copy to clipboard functionality

tile layout

@samanpwbb I'd love if you could look at applying an active class on the secondary navigation. I think I [set it up right here](https://github.com/mapbox/assembly/blob/0bef72a79864609606ece46c1cbd072f0754cfe3/site/routes.js#L209) but the value seems to get mutated by the time it [reaches here](https://github.com/mapbox/assembly/blob/0bef72a79864609606ece46c1cbd072f0754cfe3/site/navigation.js#L8)?

- Closes #587
- Closes #157
- Closes #145
- Closes #261
- Closes #442


site/examples/index.js
```diff
@@ -0,0 +1,29 @@
...
+          >{child.name}</a>
+        </div>
+      );
+    }
+
+    return (
+      <div>
+        <div className='my24'>
+          <h1 className='txt-subhead txt-bold my24'>Examples</h1>
+          <p>Assembly isn't useful until you know how to construct the markup with class names to get the design you want. The links below feature a variety of snippets that cover common component and layout interfaces.</p>
```
samanpwbb: I would rephrase this to be more positive? Starting a sentence with "Assembly isn't useful..." seems weird.

site/routes.js
```diff
@@ -23,39 +31,74 @@ import { orderSections } from './order-sections';
...
+    children: [{
+      name: 'Inputs',
+      route: '/examples/inputs/',
+      component: ExampleInputs
+    }, {
+      name: 'Map components',
+      route: '/examples/map-components/',
+      component: ExampleMapComponents
+    }, {
+      name: 'Miscellaneous',
```
samanpwbb: Not sure about the miscellaneous category. There is good stuff in here, I think we should have _more_ categories – will be easier for ppl to find what they are looking for: ('tooltip', 'badges', 'progress bar' as separate pages).

samanpwbb (changes requested):
I wrote a whole comment here and it never got posted!

Anyway, my main feedback:

I think this is a decent start and is close to being merge-able. I was expecting something more like the tachyons examples, where each example was full width. Unless we're using a lot of viewport units, I don't think full width is necessary. 

Additional things that can be ticketed for post-merge:

- Add toggle to view the mobile version of the site (will require iframing examples).
- Make the landing page more useful or remove it? Right now it repeats the sidebar contents. We could link right to the first example.

site/routes.js
```diff
@@ -140,34 +162,70 @@ function buildRoutes() {
...
           </Page>
         );
         break;
       }
       default:
-        console.error(r.name + ' has no matching component.');
+        if (r.children) props.children = r.children;
+        r.component = (
+          <Page navData={navData}>
+            <r.component {...props} />
```
samanpwbb, line 202: I didn't know r.component could be used like that!

tristen:
> Make the landing page more useful or remove it? Right now it repeats the sidebar contents. We could link right to the first example.

It could be more useful? I like that we can iterate there that Assembly is atomic and examples help illustrate how this power comes to life.

> I think this is a decent start and is close to being merge-able. I was expecting something more like the tachyons examples, where each example was full width. Unless we're using a lot of viewport units, I don't think full width is necessary.

Originally I set out for that kind of layout but yeah ... I'm not sure full-width is necessary atm

site/routes.js
tristen: Sounds good. Thinking `map-components` should just be `Legends` too.

dasulit (approved):
This is a great starting point, I'm excited to start using and building upon these, gonna use those legends all the time.

Agree with Saman that the sentence starting with "Assembly isn't useful until..." could be spun differently, otherwise I have nothing to add!

site/template.html
```diff
@@ -12,5 +12,7 @@
 <body>
 {content}
 </body>
+<script src='/assembly/clipboard.min.js'></script>
```
davidtheclark, line 15: When I run `npm start` and try copying, it does not work. Something is missing ...

site/documentation/entry.js
```diff
@@ -14,7 +14,7 @@ class Entry extends React.Component {
     const { props } = this;
 
     const example = !props.parsedComment.example ? null : (
-      <HtmlExample code={props.parsedComment.example.description} />
+      <HtmlExample code={props.parsedComment.example.description} nocopy={true} />
```
davidtheclark: Why not make this `copy={false}`? I find that easier to understand than the double-negative (i.e. if I want copy it should be `nocopy={false}`).

site/examples/inputs.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { HtmlExample } from '../html_example';
+
+const inputWithButton = `<fieldset class='relative'>
+  <button class='btn absolute top right w96 round-r'>Search</button>
+  <input class='input pr96' placeholder='Search'>
```
davidtheclark: The input needs some additional padding on the right, so text doesn't jam up against the button.

site/examples/map_components.js
```diff
@@ -0,0 +1,86 @@
...
+    <div class='flex-child flex-child--grow bg-blue-dark h12'></div>
+    <div class='flex-child flex-child--grow bg-red-faint h12'></div>
+    <div class='flex-child flex-child--grow bg-red-light h12'></div>
+    <div class='flex-child flex-child--grow bg-red h12'></div>
+    <div class='flex-child flex-child--grow bg-red-dark h12'></div>
+  </div>
+  <div class='flex-parent txt-xs'>
+    <div class='flex-child flex-child--grow'>Low</div>
+    <div class='flex-child flex-child--grow align-r'>High</div>
+  </div>
```
davidtheclark: This is a case where floats might be more straightforward:

```html
<div class='clearfix txt-xs'>
  <div class='fl'>Low</div>
  <div class='fr'>High</div>
</div>
```

Disagree?

site/examples/map_components.js
```diff
@@ -0,0 +1,86 @@
...
+    </div>
+  </div>
+  <div class='flex-parent txt-xs'>
+    <div class='flex-child flex-child--grow align-center'>3px</div>
+    <div class='flex-child flex-child--grow align-center'>6px</div>
+    <div class='flex-child flex-child--grow align-center'>12px</div>
+    <div class='flex-child flex-child--grow align-center'>18px</div>
+    <div class='flex-child flex-child--grow align-center'>24px</div>
+  </div>
+</div>`;
```
davidtheclark: The horizontal alignment looks off to me:

screen shot 2017-01-17 at 11 45 42 am

site/examples/miscellaneous.js
```diff
@@ -0,0 +1,76 @@
...
+  <button>
+    <svg class='icon icon--s inline-block align-middle ml3'><use xlink:href='#icon-close'></use></svg>
+  </button>
+</div>`;
+
+const progressBar = `<div class='h12 bg-darken10 relative round-full'>
+  <div class='absolute h12 bg-green-light round-full' style='width:50%;'></div>
+</div>`;
+
+const chip = `<div class='flex-parent flex-parent--center-cross flex-parent-inline relative w240'>
```
davidtheclark: Chip?! Weird.

site/examples/navigation.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import { HtmlExample } from '../html_example';
+
+const menuNavigation = `<nav'>
+  <a class='btn color-blue color-white-on-active bg-transparent bg-darken5-on-hover bg-blue-on-active txt-s ml3 is-active' href='#'>
+    <svg class='icon inline-block align-middle'><use xlink:href='#icon-home'/></svg> Home
```
davidtheclark: The vertical centering is better if you use flexbox. Or you could wrap the text in a `<div class='inline-block align-middle'>`

site/examples/navigation.js
```diff
@@ -0,0 +1,54 @@
...
+  </a>
+  <a class='btn color-blue color-white-on-active bg-transparent bg-darken5-on-hover bg-blue-on-active txt-s ml3' href='#'>
+    <svg class='icon inline-block align-middle'><use xlink:href='#icon-user'/></svg> Account
+  </a>
+  <a class='btn color-blue color-white-on-active bg-transparent bg-darken5-on-hover bg-blue-on-active txt-s ml3' href='#'>
+    <svg class='icon inline-block align-middle'><use xlink:href='#icon-logout'/></svg> Logout
+  </a>
+</nav>`;
+
+const underlinedTabs = `<div class='bg-gray border-b border--black flex-parent flex-parent--stretch-cross h40 txt-bold txt-s'>
```
davidtheclark: Because the body background is so dark, it's hard to see what's going on here — that the white border is overlapping a black border.

site/examples/navigation.js
```diff
@@ -0,0 +1,54 @@
...
+  <button class='flex-child mb-neg1 px12 py6 border-b border--black border--white-on-active color-lighten50 color-white-on-active color-lighten75-on-hover'>One</button>
+  <button class='flex-child mb-neg1 px12 py6 border-b border--black border--white-on-active color-lighten50 color-white-on-active color-lighten75-on-hover is-active'>Active</button>
+  <button class='flex-child mb-neg1 px12 py6 border-b border--black border--white-on-active color-lighten50 color-white-on-active color-lighten75-on-hover'>Two</button>
+</div>
+<div class='bg-gray-dark color-white p12'>
+  Changing content
+</div>`;
+
+const breadcrumbs = `<div class='flex-parent flex-parent--stretch-cross txt-s'>
```
davidtheclark: What does `flex-parent--stretch-cross` contribute?

site/examples/navigation.js
```diff
@@ -0,0 +1,54 @@
...
+  <a href='#' class='flex-child link link--darken50'>One</a>
+  <svg class='icon flex-child align-middle'><use xlink:href='#icon-caret-right'/></svg>
```
davidtheclark: `align-middle` should be tossed out in favor of flexbox vertical centering.

site/examples/sidebar_apps.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { HtmlExample } from '../html_example';
+
+const basicSidebar = `<div class='viewport-full relative clip'>
+  <div class='bg-darken10 viewport-twothirds viewport-full-ml ml0 ml300-ml absolute top left right bottom'></div>
```
davidtheclark: Why use absolute positioning and an offsetting margin instead of flexbox?

davidtheclark (commented):
Coo-oo-ool @tristen.

site/examples/map_components.js
tristen: Yes `clearfix` sux and is a hack.

site/examples/miscellaneous.js
tristen: This isn't a chip :P @samanpwbb @dasulit @davidtheclark is there a better name for this? I'm also happy to remove it.

screen shot 2017-01-17 at 2 02 03 pm

site/examples/sidebar_apps.js
tristen: Ahh good catch. I should use flexbox here.

site/examples/inputs.js
tristen: hmm? are you seeing this spacing as tight?

screen shot 2017-01-17 at 2 06 52 pm

I wanted to avoid adding too many additional classes.

davidtheclark: If you type such that you fill up that input, the text is going to be jammed up against the button, right?

Yes, we should avoid *unnecessary* classes; but if this is a pattern that we want people to copy-paste and use all the time, we should not hold back from adding the classes needed to get it right.

site/examples/map_components.js
davidtheclark: What you've got there seems kind of hacky.

site/examples/navigation.js
tristen: @davidtheclark removing `align-middle` sounds good.

> `align-middle` should be tossed out in favor of flexbox vertical centering.

Isn't that what `flex-parent--center-cross` contributes?

site/examples/inputs.js
samanpwbb: Could this example be rewritten with input that has: `border--r--0 round-left` and is positioned beside a button, with no absolute positioning? I think that would be a cleaner implementation.

site/examples/navigation.js
davidtheclark: You have *stretch*-cross. Is that a typo?

site/examples/miscellaneous.js
samanpwbb: Hm I don't know what that thing is called, but I do know we have them in mapbox studio!

screen shot 2017-01-17 at 2 22 49 pm

site/template.html
tristen: @davidtheclark can you post a bit more info? copy works for me and it seems to be added to the `dist` directory.

davidtheclark: 🤔 No more information really. I checked out this branch, did `yarn install`, then ran it, and it didn't work. That `clipboard.min.js` file was not being found by the browser. Are you sure everything's committed & building correctly? Could be that it got set up correctly for you but not reproducibly? Is it working for others?

site/examples/map_components.js
tristen: I'm not changing this to floats. What else you got?

site/examples/navigation.js
tristen: 0_______0

PR #639, opened 2017-01-19
Fix flex-child--grow so element does not overflow

davidtheclark:
Important fix!

Currently, a `flex-child--grow` element with long content can overflow its parent:

screen shot 2017-01-19 at 9 14 44 am

This fixes that.

@samanpwbb for review.

### DIFF
--- a/site/catalog/flexbox.js
+++ b/site/catalog/flexbox.js
@@ -1,0 +1,63 @@
+import React from 'react';
+
+class Flexbox extends React.Component {
+  render() {
+    return (
+      <div>
+        <h2 className='border-b border--2 border--gray-faint pb6 mt72 mb24 txt-l txt-uppercase txt-bold'>
+          Flexbox
+        </h2>
+
+        <div className='mb24'>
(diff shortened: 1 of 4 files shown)


dasulit:
I've been constantly running into this while using Assembly in another project and thought I was just using flexbox wrong X_X

thanks for catching!

davidtheclark:
@samanpwbb I went further with this. I removed `flex-shrink: 0` from `flex-child` because I think it was producing counter-intuitive results. I made a little debug section to illustrate. Here's what we have currently (before this fix):

screen shot 2017-01-19 at 9 45 26 am

I think it is confusing that in the first example, where `flex-child` is used on both children without any modifiers, the children grow indefinitely according to their content. I think better this screenshot after the fix is preferable:

screen shot 2017-01-19 at 9 47 49 am

On review of the relevant [flexbugs](https://github.com/philipwalton/flexbugs), I don't think that we are going to introduce any possible problems on browsers other than IE10. So I think this is ok. Do you agree?

samanpwbb:
This looks solid. 

> other than IE10. So I think this is ok. Do you agree?

I agree. We are ie 11+. For some reason I thought flex-child was addressing ie11/edge bugs...

PR #651, opened 2017-01-19
Tooltip examples

sarahkleins:
Adds more tooltip examples:

- Tooltip with close button
- Basic tooltip w/left triangle
- Basic tooltip w/right triangle

### DIFF
--- a/site/examples/tooltips.js
+++ b/site/examples/tooltips.js
@@ -6,6 +6,17 @@ const basic = `<div class='flex-parent-inline flex-parent--center-main flex-pare
   <span class='triangle triangle--d'></span>
 </div>`;
 
+const basic_with_closure = 
+`<div class='flex-parent-inline flex-parent--center-main flex-parent--wrap'>
+  <div class='bg-darken5 color-darken50 align-center col col--12 p18 round txt-bold txt-s relative'>
+    <button class='absolute top right'>
+      <svg class='icon link color-darken50'><use xlink:href='#icon-close'></use></svg>
+    </button>
+    Hello world!
+  </div>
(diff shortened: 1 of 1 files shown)


samanpwbb:
Mostly looks good!

- For consistency, I would make all the examples dark (or light). 
- The left and right labels should be reversed. The one labelled (left) seems to be a tooltip that would appear to the right and vice versa.

site/examples/tooltips.js
```diff
@@ -6,6 +6,27 @@ const basic = `<div class='flex-parent-inline flex-parent--center-main flex-pare
   <span class='triangle triangle--d'></span>
 </div>`;
 
+const basic_left = `<div class='flex-parent-inline flex-parent--center-cross'>
+  <span class='triangle triangle--l'></span>
+  <div class='bg-darken75 color-white align-center col col--12 p6 round txt-bold txt-s'>Hello world!</div>
```
davidtheclark: A couple of flexbox points:

- The flex children should have `flex-child` classes.
- We shouldn't need to use `col col--12` on these divs. If the desired effect is to have the triangle take up the space it needs and the body take up the rest of the space, we can do that like so:

  ```html
  <div class='flex-child flex-child--no-shrink'><!-- triangle in here --></div>
  <div class='flex-child flex-child--grow'><!-- body in here --></div>
  ```

  If the parent, though, is `flex-parent-inline`, I'm not sure you need to include the `--no-shrink` and `--grow` modifiers are all?

davidtheclark (approved):
👍 from me. @samanpwbb did you have any other issues or should we merge?

Thanks @sarahkleins!

PR #652, opened 2017-01-19
Full screen fixed header

sarahkleins:
Adds a fixed header example 
screenshot 2017-01-19 18 53 11


site/examples/fixed_header.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { HtmlExample } from '../html_example';
+
+const fixed_header = `<div class='bg-darken10 viewport-half'>
+<!-- In practice, it makes sense to use this for the container instead:
```
samanpwbb: The fixed and not-fixed versions of this header should be as similar as possible. You could use:

`<div class='absolute top right bottom left scroll-auto'></div>`

as an alternative to 

`<div class='fixed top right bottom left scroll-auto'></div>`

Only one class different, and should communicate the idea just as well.

---

Also I don't think you need `scroll-auto` on this example.

site/examples/fixed_header.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { HtmlExample } from '../html_example';
+
+const fixed_header = `<div class='bg-darken10 viewport-half'>
+<!-- in practice, the header container should use the class "fixed" instead of "absolute" -->
+    <div class='relative'>
```
davidtheclark, line 6: We've been using 2 spaces for indentation everywhere else.

davidtheclark (commented):
@sarahkleins I think the implementation is good here, but I'm wondering what purpose the example serves. All that it's doing right now is showing `absolute` in action. Are there certain aspects of a fixed header that you consider difficult or non-intuitive?

davidtheclark:
@sarahkleins Continuing my comment above ... Maybe it needs some long, scrolling content (which also has a top margin to offset the heading)? Then we could see the stickiness in action.

sarahkleins:
@tristen @davidtheclark This example was at @tristen 's suggestion to replace this guy: http://bl.ocks.org/tristen/768ac555d78b7b7123791580b2f21b90

I can also add more content, like a long scrolling sidebar offset w/the header? Thoughts @tristen ?

tristen:
Not sure how important this layout is @sarahkleins. I don't think there should be any scrollable content on the main container as it was designed to be a full page app. The origin of this layout came from a support request(s) where a user wanted a full page map and header with no overflow (cc @geografa as I think we talked about it in the past?)

sarahkleins:
Checking up on this - any consensus on if we need this example? If not, I can close the PR.

cc @tristen @samanpwbb @davidtheclark

tristen:
@sarahkleins Let's maybe pause on this one? If it's a common layout as we start to use Assembly more we can return to it.

Issue #626, opened 2017-01-16
Show code for each class, collapsed by default


tmcw:
To make it obvious what rules you get with each class, we should show the source in the documentation.

davidtheclark:
Closed by #679, which provides links to source code on GitHub.

PR #660, opened 2017-01-20
Exposing users to CSS

davidtheclark:
I was thinking about ways to address #626 and expose users to the CSS code.

This implements one possibility: just add a little "Show CSS" button.

screen shot 2017-01-20 at 4 05 27 pm
screen shot 2017-01-20 at 4 05 31 pm

There is one problem with this: most selector *groups* (e.g. the widths) do not have a corresponding `referencedSource` at all, some just don't have one that makes a lot of sense. 

To address this problem, I see two options:

- Work on a way in documentation-css to better control the `referencedSource` (e.g. so it can refer to a set of rules rather than just one) and then modify our CSS to use that.
- Work on a way in documentation-css to pass along information about the source file, line, and column; then instead of an on-page CSS toggle, provide a link to the original documentation-generating comment in GitHub.

What would you prefer (cc @tmcw)?

### DIFF
--- a/site/css_snippet.js
+++ b/site/css_snippet.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import Lowlight from 'react-lowlight';
+import cssLanguage from 'highlight.js/lib/languages/css';
+
+Lowlight.registerLanguage('css', cssLanguage);
+
+class CssSnippet extends React.Component {
+  render() {
+    // flex-parent removes line-height
+    return (
+      <div>
+        <button
+          className='flex-child btn btn--xs btn--blue round'
+          data-css-snippet-button={true}
+        >
+          Show CSS
+        </button>
+        {/* The code block must be the next sibling */}
+        <div className='none mt6 pre bg-gray-faint scroll-auto hmax180 border border--gray-light round-b relative'>
+          <div className='absolute top right p12'>
+            <button
+              data-clipboard-text={this.props.code}
+              className='ml3 btn btn--xs btn--darken25 round js-clipboard'>
+              Copy
+            </button>
+          </div>
+          <Lowlight
+            language='css'
+            value={this.props.code}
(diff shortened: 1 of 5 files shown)


davidtheclark:
I worked out a proof-of-concept with links to source code on GitHub — so it can be done. The links are to places like this one: https://github.com/mapbox/assembly/blob/0.5.0/src/forms.css#L86.

What do you ya'll think would be the better solution?

samanpwbb:
I think what you have in this branch would work, but a couple big changes would help a lot:

- Rules that have side-effects like `prose--dark` should show _every rule that declares prose--dark anywhere in the rule_ – I want to see `.prose--dark blockquote`, ect, here :

screen shot 2017-01-23 at 10 55 34 am

- If a rule is comma-separated in the source, right now all the options get displayed together. It would be better filter out the rules that aren't being explicitly documented:

screen shot 2017-01-23 at 10 56 22 am

--- 

The end goal would be to make the CSS easy to copy+paste and to clarify the purpose of rules. To do this, we'd need to do a lot of manipulation of source CSS. Not sure that's worth it.

davidtheclark:
> Not sure that's worth it.

@samanpwbb Yeah, it seems like the amount of work to get good code snippets displayed on the page is much more than to get links to source code.

davidtheclark:
Realizing that the only way to have links to source code for all rules is to save the generated CSS (color variants, layout scales) to files and commit those to the repo. Any strong objections to that @tristen and @samanpwbb?

samanpwbb:
@davidtheclark I have no objections to committing generated CSS files. I think it would make this project more approachable and easier to understand.

tristen:
No strong objections to this with the condition that we have a `/* this is generated */` banner for the clarity of anyone contributing?

PR #665, opened 2017-01-21
Lint and bundle site's runtime JS

### LINKED ISSUE #663: site/js/ scripts should be linted for browser compatibility
For example, they should not use `arrow` functions.

davidtheclark:
Closes #663. Also uses browserify to compile and bundle these scripts.

@samanpwbb for review.


samanpwbb:
👍

Issue #661, opened 2017-01-20
expander.js should use event delegation


davidtheclark:
Instead of assigning a different event listener to each *See all* button, we should add one listener to the document and react only if the click is on a relevant button.

@dasulit Have you used this technique before? It's worth learning about — I'd be happy to chat with you about it. Or if you'd prefer I can do this in a PR myself.

dasulit:
@davidtheclark I'll give it a try! I have not used event delegation before, but I can look into it and take a shot at a PR on Monday.

PR #670, opened 2017-01-23
Use event delegation in expander.js

dasulit:
This PR fixes https://github.com/mapbox/assembly/issues/661 by adding an event listener on the whole documentation wrapper, rather than adding individual listeners on each button.

I got rid of some bad ID naming I had done earlier, but I don't think reading the way I'm reading the DOM in this PR is very elegant, either. @davidtheclark do you have any suggestions for a better approach to event delegation here?

### DIFF
--- a/site/js/expander.js
+++ b/site/js/expander.js
@@ -1,25 +1,14 @@
-// 	event handlers on each button
+// 	'See All' button event handlers on Documentation page
 //	on click hide elem, show all
 
-const buttons = document.querySelectorAll('button[id^="expandButton-"]');
-let i;
-
-for (i = 0; i < buttons.length; i++) {
-  buttons[i].addEventListener('click', function () {
-
-    const classId = this.id.replace('expandButton-', '');
-    const expanded = document.getElementById('expanded-' + classId);
-    const collapsed = document.getElementById('collapsed-' + classId);
-
-    if (expanded.className === 'none') {
-      expanded.className = 'inline';
-      collapsed.className = 'none';
-      this.textContent = 'See fewer';
-    } else if (expanded.className === 'inline') {
-      expanded.className = 'none';
-      collapsed.className = 'inline';
-      this.textContent = 'See all';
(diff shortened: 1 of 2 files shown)


site/js/expander.js
```diff
@@ -1,25 +1,14 @@
...
-
-    if (expanded.className === 'none') {
-      expanded.className = 'inline';
-      collapsed.className = 'none';
-      this.textContent = 'See fewer';
-    } else if (expanded.className === 'inline') {
-      expanded.className = 'none';
-      collapsed.className = 'inline';
-      this.textContent = 'See all';
+document.getElementById('Documentation').addEventListener('click', (e) => {
```
davidtheclark: You could add this listener directly to the document: `document.addEventListener('click', ..)`

site/js/expander.js
```diff
@@ -1,25 +1,14 @@
...
+  if (e.target && e.target.className.includes('expand-button')) {
```
davidtheclark: I like to separate JS selectors from CSS class names, so you could freely change the CSS without affecting the JS. I'd suggest applying a `data-*` attribute that the JS reads. E.g. if you use `data-expand-button` you could replace this line with `if (!e.target.getAttribute('data-expand-button') return`;

site/js/expander.js
```diff
@@ -1,25 +1,14 @@
...
+    if (e.target.parentNode.children[0].className === 'none') {
+      e.target.parentNode.children[0].className = 'inline';
```
davidtheclark: If you use the `classList` API here, you won't be overriding other classes that might exist on the element. See https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

davidtheclark (commented):
Looks like a good start. I left a few suggestions.

dasulit:
@davidtheclark cool, thanks for your quick feedback! I just made some amendments, let me know what you think.

davidtheclark (approved):
👍

PR #683, opened 2017-01-26
Add txt-strike

### LINKED ISSUE #681: txt-strike rule
Assembly doesn't currently have a txt-strikethrough rule - can it?

davidtheclark:
Closes #681.

Also adds `.prose strike`. While I was at it, I added some representative elements to the `.prose` example.

@samanpwbb for review.

### DIFF
--- a/src/typography.css
+++ b/src/typography.css
@@ -534,6 +534,17 @@ textarea {
 .txt-underline-on-hover:hover { text-decoration: underline !important; }
 
 /**
+ * Strikethrough text.
+ *
+ * @memberof Type utils
+ * @example
+ * <div class='txt-strike'>txt-strike</div>
+ */
+.prose del,
+.prose s,
+.txt-strike { text-decoration: line-through !important; }
+
+/**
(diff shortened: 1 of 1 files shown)


samanpwbb:
Is the `strike` element the right thing to use? What do most markdown parsers output? I notice that github's  uses the [`del`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/del) element.

davidtheclark:
Oh sheesh you're right. Looks like `<del>` is appropriate for deleted text, `<s>` for other text that is stricken-through. Probably we should make `.prose` address both. But then what classname — maybe stick with `txt-strike`?

samanpwbb:
> Oh sheesh you're right. Looks like `<del>` is appropriate for deleted text, `<s>` for other text that is stricken-through. Probably we should make .prose address both. But then what classname — maybe stick with txt-strike

That all makes sense 👍  – we can add `.prose` selectors for both `s` and `del`?

samanpwbb (approved):

PR #685, opened 2017-01-26
Fix grid media query classes

### LINKED ISSUE #682: grid--gut does not properly un-offset col-mm
grid--gut does a margin-left trick, and when it contains `col` classes, it does the right thing. but when it contains `col-mm` mediaqueried col classes, it doesn't, and the column ends up too far to the left.

davidtheclark:
Closes #682.

This involved some significant refactoring of `build-layout-scales.js`. I modified the `buildMediaRules` function so that it could handle the multi-selector rules that we needed. This modification also allowed us to use the same template literal for non-media-queried and media-queried rules, instead of the special argument syntax we had before.

I added an example to the Catalog that was broken, illustrating #682, and now works.

@tristen for review.


tristen:
Code looks good! Passing a template argument is great.

tristen (approved):

Issue #673, opened 2017-01-24
Toggles should not have extra padding by default


davidtheclark:
Right now toggles by default have a padding of `6px` all the way around. I'm thinking maybe we shouldn't have this. 

I'd bet toggles are more frequently used as standalone UI components than form controls, so I think the height does not need to match inputs.

Also, the x-padding means that when you don't have a border you will have to override the padding to get horizontal alignment with other things on the page.

<img width="343" alt="screen shot 2017-01-24 at 12 39 29 pm" src="https://cloud.githubusercontent.com/assets/628431/22263591/34369062-e232-11e6-84ad-cbe7b31cc1a0.png">

samanpwbb:
I hear what you are saying. There will be no way to align these little things to vertical baseline without the built-in padding. I would be fine removing the padding, but it will cause similar problems for other cases, like when you are trying to correctly center a toggle inside a form element:

![screen shot 2017-01-24 at 2 44 39 pm](https://cloud.githubusercontent.com/assets/108094/22263790/a8af6264-e243-11e6-8e21-238b7792aa48.png)

davidtheclark:
@samanpwbb: That screenshot you posted won't be helped by the padding, right? You'd need to vertically center. Regarding the vertical baseline, though, I'm happy to defer to what you want, and I can override padding with utility classes as needed.

samanpwbb:
I am indecisive. I see benefits to both. Could we remove the padding by default, and add it only for stroked toggles?

PR #688, opened 2017-01-26
simplify toggles, remove built-in padding and stroke modifiers, close…

samanpwbb:
…s #673 

I ended up removing the toggle stroke modifier, and removing all built in padding. The built in padding was there to ensure toggles are aligned when user adds background to toggles – but we can trust implementors to use padding classes to acheive the same effect. The stroke modifier class had a similar purpose: to ensure that both small and large toggles are aligned to grid even when stroked – again, we can leave this responsibility to implementors.  

@davidtheclark  for review

### DIFF
--- a/site/catalog/toggles.js
+++ b/site/catalog/toggles.js
@@ -32,7 +32,7 @@ const lightenColors = [
 
 function ToggleEl(props) {
   const toggleClass = `toggle toggle--active-${props.activeColor} toggle--${props.color}`;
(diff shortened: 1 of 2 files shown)


davidtheclark (approved):
Looks good. My only concern is that we don't have padding classes for `2px` and `4px` so can't easily get the same heights we were getting with `toggle-group--stroke`. If that's ok, then 🆗 .

samanpwbb:
I think it's fine. Toggles themselves are aligned to the baseline grid. If someone wants padding or borders around toggle container, they're on their own – can always use flexbox to center.

PR #690, opened 2017-01-26
Add all flex-direction

### LINKED ISSUE #689: Add all flex directions
Which would be:

- `flex-parent--row` (The direction `flex-parent` gives us by default)
- `flex-parent--row-reverse`
- `flex-parent--column` (Currently the only direction variation in Assembly)
- `flex-parent--column-reverse`

Reverse is great for scenarios where your markup structure looks like:

```html
<container>
 <sidebar />
 <map />
</container>
```

And nn desktop you want the order to be sidebar/map and on mobile map/sidebar:

screen shot 2017-01-26 at 12 04 06 pm

tristen:
- Closes #689

### DIFF
--- a/src/layout.css
+++ b/src/layout.css
@@ -565,6 +565,45 @@
 .flex-parent--column { flex-direction: column !important; }
 
 /**
+ * Set the direction of the main axis to bottom-to-top (default is left-to-right).
+ *
+ * @memberof Flexbox
+ * @example
+ * <div class='bg-darken10 flex-parent flex-parent--column-reverse'>
+ *  <span class='flex-child'>1</span>
+ *  <span class='flex-child'>2</span>
+ *  <span class='flex-child'>3</span>
+ * </div>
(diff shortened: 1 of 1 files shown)


src/layout.css
```diff
@@ -565,6 +565,45 @@
...
+ * <div class='bg-darken10 flex-parent flex-parent--column-reverse'>
+ *  <span class='flex-child'>1</span>
+ *  <span class='flex-child'>2</span>
+ *  <span class='flex-child'>3</span>
+ * </div>
+ */
+.flex-parent--column-reverse { flex-direction: column-reverse !important; }
+
+/**
+ * Set the direction of the main axis to left-to-right (The default value of `flex-parent`).
```
davidtheclark: No capital "T" in "the default value".

davidtheclark (approved):
One capitalization qualm, then let's merge.

PR #695, opened 2017-01-27
improve vibes between txt-mono and txt-{size}, closes #694

### LINKED ISSUE #694: txt-mono doesn't play well with txt-{scales}
I've noticed that the font-size: 90% rule on `txt-mono` fights the font-size rules on `txt-s`, `txt-xs`, etc. when paired together.

txt-mono-bug

samanpwbb:
@dasulit for review

### DIFF
--- a/src/typography.css
+++ b/src/typography.css
@@ -20,6 +20,114 @@ textarea {
 }
 
 /**
+ * Style text like a keyboard button.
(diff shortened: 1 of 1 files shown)


dasulit (approved):
good vibes 👍

PR #697, opened 2017-01-28
add transition class, closes #696

### LINKED ISSUE #696: `transition` class
we have a transition-none class https://github.com/mapbox/assembly/blob/22adf098b31267eda204d156a1a0c857e97b68e7/src/miscellaneous.css#L109-L111

it would make sense to also have a `transition` class that matches buttons and other form element's transition:

```css
.transition {
  transition: all .125s;
}
```

samanpwbb:
@tristen for review

### DIFF
--- a/src/miscellaneous.css
+++ b/src/miscellaneous.css
@@ -100,6 +100,17 @@
 }
 
 /**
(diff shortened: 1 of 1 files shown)


tristen (approved):
Looks great!

PR #700, opened 2017-01-31
button font size + padding

mayagao:
@samanpwbb @tristen make button font size slightly smaller? Should we introduce a different font size here? If not i can move this to mapbox addon.

screen shot 2017-01-31 at 11 47 28 am
screen shot 2017-01-31 at 11 47 11 am

### DIFF
--- a/src/buttons.css
+++ b/src/buttons.css
@@ -33,8 +33,9 @@
   border-style: solid;
   border-color: transparent;
   border-radius: 18px; /* fully round by default */
-  padding: 4px 12px; /* Padding accommodates borders */
+  padding: 4px 18px; /* Padding accommodates borders */
   font-weight: bold;
+  font-size: 13px;
   text-align: center;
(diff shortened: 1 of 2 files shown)


samanpwbb:
Button font size beeds to be left to txt-* classes

mayagao:
per chat let's not make the default more opinionated 🙅🏿

tristen:
@mayagao If we're always resetting the default style because it doesn't work means we should change it! Core components should function well on their own without always requiring overrides. We can still be slightly opinionated; utility classes will still easily override things.

I think we should re-open this pull request? The only change I'd make is to use the variable for the `txt-s` size over a hard `13px` value and to confirm the component still aligns to the pixel grid.

samanpwbb:
> I think we should re-open this pull request? The only change I'd make is to use the variable for the txt-s size over a hard 13px value and to confirm the component still aligns to the pixel grid.

Okay, I am on board with that 👍  – make sure it still aligns to the grid

PR #704, opened 2017-02-01
examples -- Full screen spinner

sarahkleins:
full_width_spinner

Adds a full-width loading spinner example

### DIFF
--- a/site/examples/spinner.js
+++ b/site/examples/spinner.js
@@ -1,8 +1,9 @@
 import React from 'react';
 import { HtmlExample } from '../html_example';
 
-const spinner = `<div class=' viewport-full relative clip'>
-  <div class='flex-parent flex-parent--center-cross flex-parent--center-main bg-darken10 viewport-twothirds viewport-full-ml'>
+const spinner = `<div class='viewport-full relative'>
+ <!-- the wrapper above is for documentation purposes only and not needed in practice !-->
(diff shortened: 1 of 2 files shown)


site/examples/spinner.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { HtmlExample } from '../html_example';
+
+const spinner = `<div class=' viewport-full relative clip'>
+  <div class='flex-parent flex-parent--center-cross flex-parent--center-main bg-darken10 viewport-twothirds viewport-full-ml'>
```
tristen: - I don't think `viewport-twothirds` & `viewport-full-ml` are necessary?
- The way this is set up right now, the fullscreen loading container would push the content down it should be overlaid on top of. I would give this container these classes instead: `absolute top left right bottom` to fill the entire size and be positioned outside of the document flow. I would also add the highest z-index value we provide in Assembly `z5` to ensure it's above everything.

site/examples/spinner.js
```diff
@@ -0,0 +1,24 @@
...
+    <div class='flex-child loading'></div>
```
tristen: the `flex-child` class isn't necessary here as we aren't directly changing any flex behaviour of this element.

tristen (changes requested):

site/examples/spinner.js
davidtheclark: The advertised position is that you should put `flex-child` on every child of a `flex-parent`.

tristen (approved):

PR #707, opened 2017-02-07
Scrollbars

samanpwbb:
Simple custom scrollbars:

screen shot 2017-02-07 at 4 45 33 pm

@davidtheclark for review.

### DIFF
--- a/src/miscellaneous.css
+++ b/src/miscellaneous.css
@@ -5,6 +5,46 @@
 */
 
 /**
+ * Style scrollbars. Only works in Webkit-based browsers (Safari and Chrome).
+ * On dark backgrounds, add the `scroll-styled--dark` modifier.
+ *
+ * @memberof Miscellaneous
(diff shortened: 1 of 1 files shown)


davidtheclark:
With two hearts, how can I say no?

src/miscellaneous.css
```diff
@@ -5,6 +5,46 @@
 */
 
 /**
+ * Style scrollbars. Only works in Webkit based browsers (Safari and Chrome).
+ * On dark backgrounds, add the `loading--dark` modifier.
```
davidtheclark: I don't think you want the `loading--dark` modifier.

davidtheclark (changes requested):
One doc change needed, then merge.

src/miscellaneous.css
davidtheclark: While you're at it, you could hyphenate `Webkit-based`

PR #715, opened 2017-02-13
Fix custom build colors

### LINKED ISSUE #714: Custom builds are not using user color variables
I believe this was working once, but broke as we iterated on the method of building color variants. Right now color variants are now being built to actual CSS files in `src/` that get committed; then these files are used during the build. Those built files contain the default color values. When we create a custom user build, *those files are not being rebuilt with new color values*.

davidtheclark:
Closes #714.

- `build-color-variants` now writes *variables* to the CSS (e.g. `var(--blue)`), rather than looking up those variables and writing the values. This allows it to adapt when a custom build offers different variable values.
- `build-css` now builds color variants on the fly instead of reading from `src/color-variants.css`. This allows it to use a custom build's color variant configuration and variables. `src/color-variants.css` still exists, in its default form, for documentation generation.
- All the `build-*` scripts understand a `quiet` option.
- `build-user-assets.jest` tests that custom builds are producing what we expect them to produce.

Before merging this I am going to pull in the branch and try it out for the mapbox-colors build.

@samanpwbb @tristen either of you up for reviewing this?


davidtheclark:
Looks like this is working for our mapbox-colors build!

samanpwbb (approved):
Looks good, changes make sense, ran it locally and it works 👍

PR #719, opened 2017-02-15
Better media query classes

### LINKED ISSUE #718: Media query classes not overriding like they should
Problem: If I apply `flex-parent none-mm` to an element, it never actually gets `display: none`. The reason is that in our CSS `none` and its media variants are all above `flex-parent` and its media variants. So any `flex-parent` class will always trump any `none` class.

I believe that the way to get this right is to *make media-query classes always override non-media0-query classes*. The only way I can think of to accomplish this is to move media query classes to the bottoms of stylesheets. We could do this ourselves or use a PostCSS plugin to automate it. My inclination is to do it manually, because then we are knowingly manipulating the source order of the output.

davidtheclark:
Closes #718.

@tristen for review. Is it clear enough to you how we'd add/remove media class variants?


davidtheclark:
Ugh, something seems non-deterministic about the ordering of the media queries still. I will work on fixing this.

davidtheclark:
I think I fixed the sort order.

tristen (approved):
LGTM

PR #723, opened 2017-02-16
remove invisible border frm btn and select elements, closes #701

### LINKED ISSUE #701: Remove invisible borders from non-border buttons
Right now, buttons have invisible borders: https://github.com/mapbox/assembly/blob/eafb67c644b973a6c650fff08bb01d370dfc4cd8/src/buttons.css#L32-L34

This is fine when you are using buttons unmodified, but if you are adding your own padding or text size classes to buttons, buttons get offset from pixel grid. 

I suggest we remove non-border buttons, adjust default padding, and then leave bordered-buttons the same as they are now (with borders, and with default padding adjustments to account for borders).

samanpwbb:
The effect of this PR is that buttons look and act identical to how they did before, but non-stroked buttons will now fall on the pixel grid when vertical padding is added to them.

### DIFF
--- a/src/buttons.css
+++ b/src/buttons.css
@@ -29,11 +29,8 @@
   display: inline-block;
   background-color: var(--default-primary-interactive-color);
   color: var(--white);
-  border-width: 2px;
-  border-style: solid;
-  border-color: transparent;
   border-radius: 18px; /* fully round by default */
-  padding: 4px 12px; /* Padding accommodates borders */
+  padding: 6px 12px;
   font-weight: bold;
   text-align: center;
   text-decoration: none !important; /* prevent underline when inside .prose */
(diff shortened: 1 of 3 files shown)


davidtheclark (approved):
You'll need to update jest snapshots.

Otherwise, looks fine — except one thing to consider: In the catalog examples of small pill buttons, you can see that the stroke buttons are now shorter than the non-stroke buttons. However, their vertical alignment was kind of off before because of the 1px margin between non-stroke buttons. Stupid pills. What do you think?

samanpwbb:
@davidtheclark for review again. I redid all stroked buttons with box-shadow. There's one side-effect:

screen shot 2017-02-16 at 2 18 50 pm

`btn--stroke` elements lose their colored-stroke when tab-focused. I think this is a reasonable trade-off.

PR #734, opened 2017-02-26
rework type scale, closes #731

### LINKED ISSUE #731: txt-l should have a 'headline' variant with smaller line height
This is a `txt-l txt-bold`, attempting to use it as a headline:

screen shot 2017-02-24 at 12 00 26 am

It would look better like this, with tighter line height::

screen shot 2017-02-24 at 12 00 22 am

We could add line-height modifier classes, or have separate 'body' and 'heading' typography classes, where 'heading' classes would have less line-height

samanpwbb:
This PR makes some major changes:

- renames `txt-headline` to `txt-h1`
- renames `txt-subhead` to `txt-h2`
- adds new `txt-h3`, `txt-h4`, and `txt-h5` rules that use the same font _size_ scale as the old rules, but with tighter line height. These rules match existing rules we had in place for `prose h*`.
- Changes `txt-xl` line height from 36 to 45 (if you want the old `txt-xl` line height, use `txt-h3`). 

---

In order to migrate to the version of assembly that uses these rules, users will need to find and replace txt-headline with txt-h1, txt-subhead with txt-h2, and txt-xl with txt-h3.

---

Why did I make this change? Headlines come in a variety of font sizes, so does body text. Assembly should work for sites with different 'scales' – if a designer wants [large, full-width body text](https://www.twilio.com/), they should be able to do that without writing custom CSS or making the design suffer with headline-style line height.


samanpwbb:
@tristen and @davidtheclark for review plz

davidtheclark (approved):
Love it 💓

tristen (approved):
Makes sense to me!

PR #736, opened 2017-02-26
make disabled select item look disabled in FF, closes #735

### LINKED ISSUE #735: differentiate option:disabled
Eg the text of `two` should be lighter given:

```html
<div class='select-container'>
  <select class='select'>
    <option>one</option>
    <option disabled>two</option>
  </select>
  <div class='select-arrow'></div>
</div>
```

samanpwbb:
before:
screen shot 2017-02-26 at 6 28 35 pm

after:
screen shot 2017-02-26 at 6 29 00 pm


davidtheclark (approved):
🍪 👍

PR #737, opened 2017-02-27
btn--pill-stroke proposal

### LINKED ISSUE #722: Should it be possible to combine stroked and regular buttons in pills?
Right now, adding a regular button to a pill that includes a lot of stroked buttons results in this:

screen shot 2017-02-16 at 11 57 30 am

See the 1px margin to the left of the second button? That is :no_good_man:. Fixing this in a way that keeps the class lists simple would likely be challenging.

davidtheclark:
Closes #722, if we like it.

Introduces a `btn--pill-stroke` modifier that should be used *instead of* `btn--pill` when the pill group is primarily stroke buttons. Solves #722. Adds a minuscule amount of complexity to the pill group, but not much and you already have to read the instructions.

screen shot 2017-02-27 at 10 23 52 am
screen shot 2017-02-27 at 10 24 01 am

@samanpwbb for review.

### DIFF
--- a/site/catalog/pills.js
+++ b/site/catalog/pills.js
@@ -16,10 +16,10 @@ class Pills extends React.Component {
         </div>
 
         <div className='flex-parent mb18'>
-          <button className='btn btn--red btn--stroke btn--pill btn--pill-hl round'>Confirm</button>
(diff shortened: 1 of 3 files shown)


samanpwbb (approved):
I don't think there is any possible way to do this without introducing additional classes, this makes sense to me 👍

PR #738, opened 2017-02-27
make table padding match the scale

katydecorah:
I just noticed that the padding on tr and td is 10px which no longer matches the scale. This PR updates the padding to 12px.

@samanpwbb for review


samanpwbb:
@katydecorah can you update the jest fixtures? `npm run test:jest -- -u` should work.

PR #749, opened 2017-03-05
rework sprite loading, closes #748

### LINKED ISSUE #748: don't store entire SVG data in assembly.js
assembly.js is 49kb / 15kb gzipped. That's pretty big. Most of the weight is in the form of SVG. We store full contents of every icon in assembly.js. This is unnecessary.

Let's look at the source code for one of these icons. This is the 'airplane' icon:

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<svg viewBox="0 0 18 18">
  <path d="M7,4l1.6,4H5.5c0,0-1.4-2-2.5-2H2.2L3,8l1,3h4.6L7,15h2l3.2-4H14c1,0,2-0.7,2-1.5S15,8,14,8h-1.8L9,4H7z"/>
</svg>
```

Every icon is identical, except for the value of the `d` property on path. Rather than storing full SVG strings, we could store an object with the name of the icon and the path string:

```js
{
  ...
  airplane: "M7,4l1.6,4H5.5c0,0-1.4-2-2.5-2H2.2L3,8l1,3h4.6L7,15h2l3.2-4H14c1,0,2-0.7,2-1.5S15,8,14,8h-1.8L9,4H7z"
  ...
};

```

Then, we'd have a little function that runs client side, that takes object, maps it to an SVG document template, and uses that to build the actual SVG sprite.

Questions:

- Would the process of building the SVG sprite be slow / noticeable to users if we it on the client?
- Since a lot of the data we'd be removing here is the same for each icon, would the size savings actually not be that significant (doesn't gzipping do a pretty good job collapsing redundant data)?

samanpwbb:


samanpwbb:
I have this working. Makes very minor size difference:

This branch:
```
CSS: 245 kB (minified) => 28.3 kB (gzipped)
SVG: 41.2 kB => 14.3 kB (gzipped)
```

mb-pages:
```
CSS: 245 kB (minified) => 28.3 kB (gzipped)
SVG: 48.9 kB => 14.5 kB (gzipped)
```

I did end up cleaning up the `clean-svgs` scripts, fixing a couple bad icons, and fixing a bug in the svg tests while I was at this. I think there are two decent outcomes here:

1. Merge this PR.
2. We revert changes to build-svg-loader but retain fixes to icons / tests / cleanup file.

davidtheclark:
@samanpwbb I think if there's such an insignificant size difference we should probably (?) revert `build-svg-loader`, because I think this new way is a little harder to grasp and far less common.

samanpwbb:
@davidtheclark sounds good – I think that's the right call. I am going to clean this PR up and revert changes to build-svg-loader.

PR #766, opened 2017-03-19
Test improvements

davidtheclark:
Improves a bunch of tests. Also makes the APIs of `buildCss`, `buildJs`, and `copyFonts` consistent: just one options argument.

@samanpwbb for review.


samanpwbb (approved):
Changes look good. Pulled down and ran the tests locally and they passed and all of them ran.

PR #768, opened 2017-03-20
JS icon functions

davidtheclark:
Exposes `Assembly.createIcon`, `Assembly.changeIcon`, and `Assembly.iconExists`.

Also adds ESLinting to the JS that gets exposed to the browser.

Should wait until #766 merges to update tests.

@samanpwbb for review.


samanpwbb (approved):
These basic icon functions are going to be relied on a lot, lets get them into assembly 👍

PR #771, opened 2017-03-20
Add equal-height column example

### LINKED ISSUE #751: Easy equal-height grid row
We should make it easy to create a grid row where every item has equal height. Right now you can do it by adding additional `flex` classes; but I think this use-case is common enough that we want to make an easy and bug-free implementation.

I think adding a modifier or two for `grid` classes might do the trick.

davidtheclark:
Closes #751.

I was thinking maybe the solution to #751 is just to provide an example.

@tristen for review.


tristen (approved):
LGTM!

Issue #763, opened 2017-03-15
Better strategy for vertically centering inline icons


samanpwbb:
The current implementation of icons is surprisingly difficult to work with when icons are inline. Usually we resort to a flexbox solution, which introduces a lot of extra markup and causes problems when lines wrap. Is there anything we can do to make icons work better? I feel like there must be some solution related to line heights, vertical alignment, that will make icons both center themselves on the line, and preserve their parent's overall height.

davidtheclark:
No matter what implementation we use, the icon is going to increase the line-height if it is taller than the text, right? So if you place an icon inline with text of `font-size: 12px`, that icon needs to be `height: 12px` or the line-height will increase. (Or `font-size: 12px` if using an icon font.)

I believe we'd have this same situation if we had an icon font. The difference would be that the icon font would by default inherit the font-size of the surrounding text — but that might not be what we want, either (we may want a limited number of consistent icon sizes).

Do you have examples of a system that does *not* have this issue? I guess I don't feel the same pain maybe because I don't know what ideal this falls short of.

samanpwbb:
Line height is 18, pixel height of svg is 18, but this container div ends up being 20px tall:

<img width="843" alt="screen shot 2017-05-22 at 10 34 00 pm" src="https://cloud.githubusercontent.com/assets/108094/26336130/c8b7391c-3f3e-11e7-879f-1851ab45b195.png">

samanpwbb:
The approach introduced by https://github.com/mapbox/assembly/pull/781 is not working. The line-height context of the parent of the icon ends up pushing icons too high or too low. Icon alignment is _usually_ off. We need icons to be perfectly centered:

Icon is too low:
<img width="108" alt="screen shot 2017-09-23 at 2 09 56 pm" src="https://user-images.githubusercontent.com/108094/30775798-f7d886ca-a068-11e7-819f-8b8073a16545.png">

Icon is too high:
<img width="105" alt="screen shot 2017-09-23 at 2 09 38 pm" src="https://user-images.githubusercontent.com/108094/30775803-03300354-a069-11e7-888c-5b67dccbd3a0.png">

Can we build a better `icon-inliner` class?

davidtheclark:
Isn't this problem completely solved by using flexbox? 

I guess more experimenting could come up with something that feels easier — but the safe bet is to write elements that you want vertically centered as peer flex children, right?

samanpwbb:
[...]

PR #779, opened 2017-05-24
Icon alignment

samanpwbb:
I tried one approach for #763. This works okay for a small subset of font sizes. I think we either:

1. _need_ icon modifiers for a variety of font sizes.
2. write nested selectors (`.txt-s .icon`), which is gross.
3. Try to use an `icon-aligner class on a wrapper element around the icon, and then center the icon within the wrapper.

@tristen @dasulit do either of you have feeling about this?


tristen:
I'm a little confused. The options you propose above are different from what you've added in https://github.com/mapbox/assembly/pull/779/files#diff-9d501d666b053693f59c5ba3d899debb. What works here vs what doesn't?

samanpwbb:
@tristen i don't think the code in this PR is good, but I wanted to show what I tried. It is painful to try to center SVG icons inline. Need to use align-top in order to keep text line-height consistent, but align-top... aligns icons to the top.

PR #781, opened 2017-05-24
adjust icon sizing, add icon-wrapper class, closes #763

### LINKED ISSUE #763: Better strategy for vertically centering inline icons

samanpwbb:
This PR adds two new classes for centering icons inline, plus a new catalog entry for icons. The solution uses a wrapping element, which has some advantages:

- Possible to center icons without dedicated classes for every font size
- Even if icons are larger than line height, line heights will not be effected. 
- Super convenient to use with React. 

The one big disadvantage being that it requires more markup. I don't think there's any other option though (trust me, I tried a lot of things here!)

The PR also increases the size of the small icon. 12px was just way too tiny. Nothing was legible.

### DIFF
--- a/src/icons.css
+++ b/src/icons.css
@@ -53,30 +53,31 @@
 }
 
 /**
- * Add to a wrapping element around an icon in order to align the icon with text and ensure icon doesn't impact text line height. If text uses a `txt-h{number}` class, then use the `icon-wrapper--heading` modifier.
+ * Add to a wrapping element around an icon in order to align the icon with text and ensure the icon doesn't unexpectedly shift text line height. Note that this rule depends on an expected text line height of 1.5 to 1.66, so it does not work with `txt-h{number}` classes. See the `icon-inliner--heading` modifier to center icons in heading text.
  *
  * @memberof Icons
  * @example
- * <p class='txt-l'><span class='icon-wrapper'><svg class='icon'><use xlink:href='#icon-clipboard'/></svg></span> Copy</p>
- * <p class='txt-xl'>Ring the <span class='icon-wrapper'><svg class='icon icon--l'><use xlink:href='#icon-bell'/></svg></span></p>
- * <p class='txt-s'>Keep your <span class='icon-wrapper'><svg class='icon icon--s'><use xlink:href='#icon-home'/></svg></span> safe from <span class='icon-wrapper'><svg class='icon icon--s'><use xlink:href='#icon-bug'/></svg></span></p>
+ * <p class='txt-l'><span class='icon-inliner'><svg class='icon'><use xlink:href='#icon-clipboard'/></svg></span> Copy</p>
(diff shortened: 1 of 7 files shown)


src/icons.css
```diff
@@ -33,21 +33,50 @@
...
+/**
+ * Add to a wrapping element around an icon in order to align the icon with text and ensure icon doesn't impact text line height. If text uses a `txt-h{number}` class, then use the `icon-wrapper--heading` modifier.
+ *
+ * @memberof Icons
+ * @example
+ * <p class='txt-l'><span class='icon-wrapper'><svg class='icon'><use xlink:href='#icon-clipboard'/></svg></span> Copy</p>
+ * <p class='txt-xl'>Ring the <span class='icon-wrapper'><svg class='icon icon--l'><use xlink:href='#icon-bell'/></svg></span></p>
+ * <p class='txt-s'>Keep your <span class='icon-wrapper'><svg class='icon icon--s'><use xlink:href='#icon-home'/></svg></span> safe from <span class='icon-wrapper'><svg class='icon icon--s'><use xlink:href='#icon-bug'/></svg></span></p>
+ **/
+.icon-wrapper {
```
tristen: Just curious, what makes the `!important` cut vs what doesn't?

tristen (approved):
This looks like a good option!

src/icons.css
samanpwbb: i don't have a good answer for you. The importanted properties are the ones they definitely shouldn't ever get overriden when using this class. But the same could be said for almost every property here. Should I important everything?

tristen: I would think everything except for the `top` property? I guess this wrapper should do what it says on the tin so everything could be important'ized

samanpwbb: I'll important all but top and I'm going to rename the class .icon-inline-wrapper

src/icons.css
```diff
@@ -33,21 +33,51 @@
...
+ * <p class='txt-xl'>Ring the <span class='icon-inliner'><svg class='icon icon--l'><use xlink:href='#icon-bell'/></svg></span></p>
+ * <p class='txt-s'>Keep your <span class='icon-inliner'><svg class='icon icon--s'><use xlink:href='#icon-home'/></svg></span> safe from <span class='icon-inliner'><svg class='icon icon--s'><use xlink:href='#icon-bug'/></svg></span></p>
+ **/
+.icon-inliner {
+  position: relative !important;
+  display: inline-flex !important;
+  vertical-align: top !important;
+  align-items: center !important;
+  justify-content: center !important;
+  height: 1em !important;
```
davidtheclark, line 70: @samanpwbb: I think that conventionally we haven't been using `!important` unless the declaration corresponds directly to the class name. Is there a reason we need to use `!important` on these declarations?

PR #783, opened 2017-05-25
rework size scale, closes #749

samanpwbb:
Reworks the size scale to be both consistent and practical. The new scale increments by 3 up to 6, increments by 6 up to 36, increments by 12 up to 60, increments by 60 to 120, increments by 120 up to 600, then has two final large sizes at 1200 and 1500.

Some notes:

- This PR adds new classes, and adds about 1 KB to the size of the gzipped CSS file.
- The new scale doesn't follow any perfect mathematical formula but it's practical and simple and more consistent and should be easier remember.
- A purely exponential scale is elegant, but breaks down in practice. For example, it's common to have fixed-width sidebars and modals. An exponential scale doesn't give you a good range of 'container' sizes to work with (jumps from 196 to 392 to 784).

Some questions:

- Might make sense to add 180 and 300 so the 10x scale matches the 1x scale more closely. 
- I am not sure we need 60 and 600. Dropping these makes the pattern clearer.


samanpwbb:
@tristen @mayagao the important changes are all in https://github.com/mapbox/assembly/pull/783/files#diff-759c953ebbf2b1a9d86f29ef197f9e30 – let me know what you think!

samanpwbb:
Made some further adjustments. The scale is now more rational than ever. _Every_ property has the same values. These values were picked because they cover small scale form tweaks (0-36), larger scale layout adjustments (60-180), and containers (240-600):

`0, 3, 6, 12, 18, 24, 30, 36, 60, 120, 180, 240, 300, 360, 600`

The pattern is: 
- increment 3 up to 6,
- increment 6 up to 36,
- increment 30 up to 60,
- increment 60 up to 360,
- increment 300 up to 600

@tristen can you give this a look again?

tristen:
@samanpwbb 👍 This change looks good to me

dasulit (approved):
I'm good with these changes, the system is neat and predictable. Only thing I feel like I might miss are the higher maxWidths, but otherwise 👍

samanpwbb:
@dasulit `limiter` should get you most of what you need. Note that limiter is not `!important`d on purpose – so you can mix it with other rules if you still want that max width but want more control. You can always just write custom CSS too 🙊 .

mayagao:
🎉

Issue #775, opened 2017-05-08
bold inline or inline-block text inside another inline/inline-block element offsets line height


samanpwbb:
This is a sad off-by-one situation that I would love to avoid:

![screen shot 2017-05-08 at 1 42 45 pm](https://cloud.githubusercontent.com/assets/108094/25817163/79efe87e-33f4-11e7-93cb-e180da99bc65.png)
![screen shot 2017-05-08 at 1 42 52 pm](https://cloud.githubusercontent.com/assets/108094/25817166/7b5c662e-33f4-11e7-9fb6-4cd8f3f24abe.png)

davidtheclark:
I noticed that and wished it was an optical illusion.

davidtheclark:
Reopening this following discussion at https://github.com/mapbox/assembly/pull/784#commitcomment-22307459

A couple of notes about this issue:
- It's not consistent across browsers. Firefox does not show the problem.
- There's a possibility this problem could be fixed by using a different version of Open Sans. I ran across a GitHub commenting suggesting as much: https://github.com/FontFaceKit/open-sans/issues/25
- The most immediate fix for the initial screenshotted problem is to set the links to `block` — so that's a strategy we should keep in mind in the future.

PR #784, opened 2017-05-25
force txt-bold to be 1em line height, fixes line height offsets, clos…

samanpwbb:
…es #775 

@tristen for review. I also moved `txt-bold` _above_ font declarations so something with `txt-bold txt-h2` will get the correct line height.


tristen (approved):
🔩

PR #786, opened 2017-05-30
New copy of Open Sans

### LINKED ISSUE #775: bold inline or inline-block text inside another inline/inline-block element offsets line height

davidtheclark:
Reverts prior line-height change to bold weight and, oddly enough, fixes #775 by getting a different copy of Open Sans.

I downloaded TTFs from Google Fonts, then used Font Squirrel's Webfont Generator to create `woff` and `woff2` formats. It seems that this new version does not have the same problem with line-height on bold text!

Before (notice the different heights on colored blocks and the 1px offset of the line following the bold text):

screen shot 2017-05-29 at 7 38 13 pm

After:

screen shot 2017-05-29 at 7 38 20 pm

@samanpwbb: Does this indeed fix all the problems you noticed with this bold text?


samanpwbb:
Yes, this appears to fix the problem. So glad it works! Thanks is for giving this a shot

samanpwbb (approved):

PR #788, opened 2017-06-04
reset abbr text-decoration, fixes txt-abbr rule

samanpwbb:
Minor fix.

Before:
screen shot 2017-06-03 at 11 48 23 pm

After:
screen shot 2017-06-03 at 11 48 15 pm

Question:

Should we even have a `txt-abbr` rule? I don't predict this getting much use, and it's annoying to have to maintain.

### DIFF
--- a/src/reset.css
+++ b/src/reset.css
@@ -79,6 +79,10 @@ button::-moz-focus-inner {
   border: 0;
 }
 
+abbr {
+  text-decoration: none;
+}
+
 svg {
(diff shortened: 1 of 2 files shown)


tristen:
> Should we even have a `txt-abbr` rule? I don't predict this getting much use, and it's annoying to have to maintain.

This is hardly any maintenance burden :P and it's a good thing to add to our reset!

Issue #791, opened 2017-06-06
A defense for faint gray-faint


dasulit:
[This commit](https://github.com/mapbox/assembly/commit/11fc2205c3222467e9f9f85990b4c873e0b1322c) flew by weekend, and before Assembly 1.0 goes down I want to throw in a nitpicky defense for lightening up gray-faint again.

@samanpwbb your commit message reads "make faint gray legible"— can you say more about what you mean by legible? I don't think gray-faint should ever be used for text, but having an airy, almost-not-quite-white is really useful for backgrounds.

Here's what the new `#e5e5e5` looks like as the background for code snippets in the Assembly docs, where we're using `bg-gray-faint`... IMO this actually makes the text harder to read:

![screen shot 2017-06-06 at 4 14 49](https://user-images.githubusercontent.com/8495845/26850260-dad778b8-4ad4-11e7-855d-d28ce3d20402.png)

Maybe we could delete `color-X-faint` classes, maybe we could add a `almost-not-quite-white` class, maybe we could find a happy middle ground gray. I know this seems minor, and there were so many good changes in the weight loss program, I do use `bg-gray-faint` a lot though!

samanpwbb:
I agree. It's too dark for how it's being used. On lower-quality monitors it'll be illegible if we make it fainter, but that shouldn't cause serious problems. 

How do you feel about shifting it down to rgba(0,0,0,.05) / #f2f2f2 from rgba(0,0,0,.1) / #e8e8e8?

dasulit:
I would be happy with `#f2f2f2` 👍

samanpwbb:
@dasulit assigned this to you!

PR #792, opened 2017-06-07
Lighten gray-faint

dasulit:
Closes https://github.com/mapbox/assembly/issues/791.

I realize I was mistaken in that comment about using `gray-faint` for code backgrounds— those elements use `darken10`, so my concerns are not resolved by just lightening gray faint:

screen shot 2017-06-07 at 12 10 30

@samanpwbb is there a reason not to use `bg-gray-faint` (keeping `bg-lighten10` for the dark theme) for code elements?

### DIFF
--- a/scripts/build-color-variants.js
+++ b/scripts/build-color-variants.js
@@ -47,11 +47,13 @@ const allColors = [
   'purple-light',
   'purple-faint',
 
+  'darken5',
   'darken10',
   'darken25',
   'darken50',
   'darken75',
 
+  'lighten5',
   'lighten10',
   'lighten25',
   'lighten50',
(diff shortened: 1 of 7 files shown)


samanpwbb:
> is there a reason not to use bg-gray-faint (keeping bg-lighten10 for the dark theme) for code elements?

For a little more flexibility. By using a semitransparent background, the same effect will work across a range of backgrounds. 

It seems like the best option is to re-introduce darken5 and lighten5, but for _background_ colors only.

samanpwbb (approved):
Works, looks good. One recommendation: consider renaming the validator functions. `isNotAccessibleForBg` and `isNotAccessible` don't quite do what they sound like they should do.

dasulit:
@samanpwbb You reviewed so fast, I have a few questions!

* Is it incorrect to return null rather than `color === 'black'` for the [accessibility functions](https://github.com/mapbox/assembly/blob/5ab308be0c44e4b4724cce66a59188646b7e4925/scripts/build-color-variants.js#L77-L85)?
* The changelog says: "-faint or -light form elements are no longer available for _any element_," but it looks like we're [actually excluding -faint and -dark (and now darken5/lighten5)](https://github.com/mapbox/assembly/blob/5ab308be0c44e4b4724cce66a59188646b7e4925/scripts/build-color-variants.js#L79) from forms?
* I feel weird about [this line](https://github.com/mapbox/assembly/blob/5ab308be0c44e4b4724cce66a59188646b7e4925/scripts/build-color-variants.js#L568) (and other places following that pattern), am I doing something wrong? Why was the existing check for _not not_ accessible?

dasulit:
My questions have been answered by renaming the function names. Just committed that change:

`isNotAccessibleForBg` => `isNotAccessibleExceptButtons`
`isNotAccessible` => `isNotAccessibleExceptBg`

PR #794, opened 2017-06-10
Add icons option to build script

### LINKED ISSUE #782: Configure which icons should be included in the assembly build
We have a lot of icons, covering a wide range of use-cases, consuming `48.9 kB` / `14.5 kB` (gzipped) bandwidth. Would be nice if users of Assembly could only include some of them.

samanpwbb:
options.icons is an array of icon names. If this option is used, only icon names that match options.icons will be included in Assembly. Closes #782 

In order to properly test changes, I had to upgrade Jest (https://github.com/facebook/jest/pull/3068). 

@davidtheclark for review.


test/build-js.jest.js
```diff
@@ -30,4 +30,25 @@ describe('buildJs', () => {
...
+      })
+      .then(() => cleanup(tmp));
+  });
+
+  test('fails with invalid icon option', () => {
+    const tmp = `${getTmp()}/test.js`;
+    return expect(buildJs({
+      outfile: tmp,
+      icons: ['airplane', 'pizza']
+    })).rejects.toEqual(Error('an icon matching pizza does not exist'));
```
davidtheclark, line 52: I did not know about this `rejects` thing!

davidtheclark (approved):
Looks great!

test/build-js.jest.js
samanpwbb: Ya, I didn't either. Had trouble testing a rejecting promise case, so ended up looking into it a little and came across this. It's new as of Jest 0.20.0.

PR #795, opened 2017-06-11
add unprose class, closes #793

### LINKED ISSUE #793: unprose?
Kind of a weird idea, but maybe it is actually a good one ...

I'm thinking about situations when you're writing long prose *and* you want to intersperse special stuff inside it (e.g. UI components, custom-styled links and headings, a button, etc.). Example:

```html
<div class="prose">
  <h1>Title</h1>
  <p>Here is my <strong>prose</strong>.</p>
  <h2 class="txt-underline txt-s">Special-styled heading</h2>
  <p>I want this to be a special-styled link: <a href="place" class="bg-blue color-red">go to place</a>.</p>
  <div class="mt24">
    <a href="other-place" class="btn">Button link!</a>
  </div>
  <p>Here is more <em>prose</em>.</p>
</div>
```

The way that `prose` currently works, you would need to do one of two things:
- Ensure that you override all the `prose` styling you want to. This might be ok much of the time, but sometimes Assembly might not have the right classes for such overriding, and it ends up putting you in a slightly ambiguous situation: the class list no longer clearly explains why the element looks the way it does.
- Divide the prose block into pieces, so you don't end up with your special stuff wrapped in `prose`.

```html
<div class="prose">
  <h1>Title</h1>
  <p>Here is my <strong>prose</strong>.</p>
</div>
<h2 class="txt-underline txt-s">Special-styled heading</h2>
<div class="prose">
  <p>I want this to be a special-styled link: <a href="place" class="bg-blue color-red">go to place</a>.</p>
</div>
<div class="mt24">
  <a href="other-place" class="btn">Button link!</a>
</div>
<div class="prose">
  <p>Here is more <em>prose</em>.</p>
</div>
```

Not only is that not ideal, it also doesn't fully work: the `go to place` link is still going to have to deal with `prose` styling on the link.

So, the idea is to provide an `unprose` class, and append to all `prose` element selectors a `:not` expression that excludes those. E.g. `.prose a:not(.unprose)`, `.prose h1:not(.unprose)`.

```html
<div class="prose">
  <h1>Title</h1>
  <p>Here is my <strong>prose</strong>.</p>
  <h2 class="unprose txt-underline txt-s">Special-styled heading</h2>
  <p>I want this to be a special-styled link: <a href="place" class="bg-blue color-red unprose">go to place</a>.</p>
  <div class="mt24">
    <a href="other-place" class="unprose btn">Button link!</a>
  </div>
  <p>Here is more <em>prose</em>.</p>
</div>
```

Do you think this would be useful?

samanpwbb:
Unprose needs to be added to each element that you want to strip prose styles from, which could lead to a lot of `uprose's` all over the place. But I don't see a better way. Note that this PR also removes the line:

```css
  text-decoration: none !important; /* prevent underline when inside .prose */
```
from toggles and buttons – instead we should recommend folks use `.unprose` to solve this problem.

@davidtheclark for review.


src/tables.css
```diff
@@ -31,7 +31,7 @@
  *
  */
 .table,
-.prose table {
+.prose table:not(.unprose) {
```
davidtheclark, line 34: I think you can create unprose containers, as well, by changing these selectors like so:

`prose table:not(.unprose, .unprose table)`

I think we should still recommend applying it *directly* to the elements that you're unprosing, because otherwise you could get yourself in a little bit of a confusing tangle about whether you're in a prose section or not. But do you think it's still worthwhile to add this capability?

samanpwbb: Whoa the :not select doesn't work the way I thought it worked, but that makes sense. I prefer the idea of unprose containers, because it mirrors prose containers. Would you be opposed to using only `prose table:not(.unprose table)`?

davidtheclark: 🤔  ... We'd need to make it clear that you can't nest these things (e.g. prose > unprose > prose > unprose). That's one advantage of applying unprose directly to elements: you can't mess it up. Other reason I lean that way is that I suspect this will *mostly* be used for single elements, right? You have a special link or header, but you still want everything around it to behave as prose. You disagree?

samanpwbb: I would like to _either_ offer `.unprose` containers, _or_ offer `.unprose` as a single-element modifier. Giving the class both qualities feels unnecessary. What do you think about leaving this PR as-is and merging?

davidtheclark: I'm good with this (apply `unprose` directly to elements), as long as there's a little more explanation in the docs. However, if you really favor the container, I'm ok with you making the choice.

davidtheclark:
@samanpwbb: Before I update this branch and merge I just want to make sure we're on the same page, since it's been a little while. Are you also ok with the `.unprose` class working as it currently does?

samanpwbb:
@davidtheclark yep, current implementation is 👍

PR #802, opened 2017-07-02
Fix Browserslist

davidtheclark:
Oh geez. The Browserslist value was no good, and that resulted in old versions of Safari not getting the autoprefixing that they need.

Before on iOS 8:

screen shot 2017-07-02 at 2 51 05 pm

After on iOS 8:

screen shot 2017-07-02 at 2 52 12 pm


davidtheclark:
@tristen for review. 

@tristen and @samanpwbb: This is an important fix. We should get a release out. There are a few breaking changes in the [changelog](https://github.com/mapbox/assembly/blob/dev-pages/CHANGELOG.md#latest), though, that will make it hard to upgrade things to this necessary new version without also putting a bunch of work into migrating existing pages.

What do you think we should do? Should we get all these changes out and get pages updated on Monday? Or should we pull some Git trickery to get just *this* fix out and save the rest for later?

samanpwbb:
Great catch. I would opt for hotfixing this outside of the typical release process since we have a lot of breaking chances on dev pages now.

davidtheclark:
Going to try to hotfix this into mb-pages. Wish me luck.

davidtheclark:
Ok, this thing is deployed and dev-pages is updated with conflicts fixed.

PR #820, opened 2017-07-10
Add align-b class

samanpwbb:
Adds a class that our internal component library relies on. @davidtheclark for review.


davidtheclark (approved):
🍻

PR #821, opened 2017-07-12
Rework default behavior of col class

### LINKED ISSUE #819: easy way to make flexible, equal width children elements
There should be a way in Assembly to make an  equal-width flexible grid with any number of children (if two children, they're both 50% width, if 3 children, they're all 33.333% width, ect). Flexbox makes this easy. We could add a utility:

```css
.flex-child--equal {
  flex-basis: 0;
  flex-grow: 1;
  flex-shrink: 1;
}
```

_Otherwise_, and I think this is the better idea, we could make the `.col` class take on this behavior by default. Thoughts @mapbox/frontend ?

samanpwbb:
col class now defaults to proportional width, flexible column styling. Modifier is now optional. Closes #819

I am not going to consider this a breaking change. Is everyone okay with that decision? Although, I guess, according to semver, that distinction doesn't matter yet.

### DIFF
--- a/site/catalog/grids.js
+++ b/site/catalog/grids.js
@@ -69,7 +69,14 @@ class Grids extends React.Component {
           Grid
         </h2>
 
-        <h3 className='mb12 txt-m txt-bold color-darken50 txt-uppercase txt-s'>Standard grid</h3>
+        <h3 className='mb12 txt-m txt-bold color-darken50 txt-uppercase txt-s'>Uncontrolled grid</h3>
+        <div className='txt-s grid mb18'>
+          {[...Array(12)].map((i) => <div key={i} className='col'>
+            <div className='border px3 py3'>col</div>
+          </div>)}
+        </div>
+
+        <h3 className='mb12 txt-m txt-bold color-darken50 txt-uppercase txt-s'>Standard controlled grid</h3>
         {GridEls}
 
         <h3 className='mb12 mt60 txt-m txt-bold color-darken50 txt-uppercase txt-s'>Grid with gutters</h3>
--- a/src/layout.css
+++ b/src/layout.css
@@ -46,47 +46,69 @@
  */
 
 /**
- * Grids are made by adding the `grid` class to a parent and `col col--{size}` classes to children. Column classes include `*-mm`, `*-ml`, and `*-mxl` variations to target screen sizes.
+ * Grids require at minimum two elements: A parent with the class `grid` and a child with the class `col`. By default, each `col` in a grid will occupy equal space. The `grid` class includes `*-mm`, `*-ml`, and `*-mxl` variations to target screen sizes.
  *
- * Grids and their columns use margin and padding for layout, so you should not apply any other margin or padding to these elements. Instead, apply such styling to parent and child elements, and use these elements for the grid layout only.
+ * When using these classes, do not apply left or right margin or padding directly to the \`grid\` or \`col\` elements or you will face unexpected side-effects.
  *
  * @group
  * @memberof Grid
  * @example
  * <div class='grid'>
- *   <div class='col col--2 border border--darken10'>col--2</div>
- *   <div class='col col--4 border border--darken10'>col--4</div>
- *   <div class='col col--6 border border--darken10'>col--6</div>
+ *   <div class='col border border--darken10'>one</div>
+ *   <div class='col border border--darken10'>two</div>
+ *   <div class='col border border--darken10'>three</div>
  * </div>
  */
 .grid {
   display: flex !important;
   flex-wrap: wrap !important;
 }
-
 .col {
-  display: block !important;
+  display: block;
   max-width: 100%;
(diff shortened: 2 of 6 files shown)


site/catalog/grids.js
```diff
@@ -69,7 +69,14 @@ class Grids extends React.Component {
           Grid
         </h2>
 
-        <h3 className='mb12 txt-m txt-bold color-darken50 txt-uppercase txt-s'>Standard grid</h3>
+        <h3 className='mb12 txt-m txt-bold color-darken50 txt-uppercase txt-s'>Uncontrolled grid</h3>
+        <div className='txt-s grid mb18'>
+          {[...Array(12)].map(() => <div className='col'>
```
davidtheclark: `Warning: Each child in an array or iterator should have a unique "key" prop.`

src/layout.css
```diff
@@ -46,48 +46,67 @@
  */
 
 /**
- * Grids are made by adding the `grid` class to a parent and `col col--{size}` classes to children. Column classes include `*-mm`, `*-ml`, and `*-mxl` variations to target screen sizes.
+ * Grids require at minimum two elements: A parent with the class `grid` and a child with the class `col`. By default, each `col` in a grid will occupy equal space. The `grid` class includes `*-mm`, `*-ml`, and `*-mxl` variations to target screen sizes.
  *
- * Grids and their columns use margin and padding for layout, so you should not apply any other margin or padding to these elements. Instead, apply such styling to parent and child elements, and use these elements for the grid layout only.
+ * When using these classes, do not apply left or right margin or padding directly to the \`grid\` or \`col\` elements or you will face unexpected side-effects.
  *
```
davidtheclark, line 52: Here's a common situation: you want each of three blocks to be full-width, vertically stacked, in narrow viewports, then to be even-width columns in wider viewports. If we want to allow this to work without specifying a column number for those wider viewports, then I think we might need another class, like `col--even` which will have media variants: `col col--12 col--even-mm`?

src/layout.css
```diff
@@ -46,48 +46,67 @@
...
  * @group
  * @memberof Grid
  * @example
  * <div class='grid'>
- *   <div class='col col--2 border border--darken10'>col--2</div>
- *   <div class='col col--4 border border--darken10'>col--4</div>
- *   <div class='col col--6 border border--darken10'>col--6</div>
+ *   <div class='col border border--darken10'>one</div>
+ *   <div class='col border border--darken10'>two</div>
+ *   <div class='col border border--darken10'>three</div>
```
davidtheclark, line 59: How does this work if not *all* the columns in the grid are bare `col`s? What if you have `col`, `col col--3`, `col col--2`?

samanpwbb: It works well! I had to rewrite the col--{x} rules a bit to make it work though. There's a example of this in the col--{x} case.

davidtheclark: We should document this behavior — I don't think I saw it in the docs?

src/layout.css
samanpwbb: In that case, introducing the grid only at `-mm` would also be an acceptable solution: `col-mm` would do the trick. If you needed guaranteed full width at small sizes, `w-full w-auto-mm col-mm` would work. If other people like the feel of `col col--12 col--even-mm` I'd be okay with adding it though

src/layout.css
samanpwbb: Updated the documentation!

src/layout.css
samanpwbb: cc: @mapbox/frontend – any opinions on this?

jfurrow: I don't feel super strongly, but it seems intuitive to me to use the `col-mm` class to achieve this behavior ("this element will be a column in medium viewports").

dasulit: ➕ to using media queries for col

src/layout.css
```diff
@@ -46,48 +46,70 @@
...
-.col--7 { width: 58.3333% !important; }
-.col--8 { width: 66.6666% !important; }
-.col--9 { width: 75% !important; }
-.col--10 { width: 83.3333% !important; }
-.col--11 { width: 91.6666% !important; }
-.col--12 { width: 100% !important; }
+/**
+ * Specify a percentage-based width for a `col` in a grid. `col` elements that do _not_ have `col--{size}`
+ * modifiers but are siblings of `col`s that do have modifiers, will retain their default behavior by proportionally
+ * occuping the remaining space.  occupying the remaining space. `col--{size}` classes include `*-mm`, `*-ml`, and `*-mxl` variations
```
dasulit: `occupying the remaining space` is repeated! don't forget the `y`

src/layout.css
samanpwbb: Added col--auto for overriding col--{size} classes 👍

samanpwbb:
@mapbox/frontend how about one more final review here before I merge?

davidtheclark (approved):
👍

PR #823, opened 2017-07-13
match hover variant options to regular variant options

samanpwbb:
@dasulit we had some faulty logic in assembly that was generating the wrong color variants for hoverBorder, hoverColor, and hoverBackground.

This branch is a little larger, but this is expected and good:
CSS: 185 kB (minified) => 22.9 kB (gzipped)

dev-pages:
CSS: 176 kB (minified) => 22.2 kB (gzipped)


dasulit (approved):
Based on the additions in color-variants.css, this looks good. Thanks for fixing 😓

PR #827, opened 2017-07-26
Migrate to Batfish 🦇🐠

jfurrow:
This PR removes all of the static site building logic from the project and implement [Batfish](https://github.com/mapbox/batfish) instead. As far as I can tell, the documentation website works just as it did before. Also the the package's script names should result in the same behavior.

Along the way I had to upgrade React due to a mismatched peer dependency, so I also introduced the `prop-types` package to alleviate React's warnings about using `React.PropTypes`.

Sorry for the huge PR... 3.2k additional lines are from `yarn.lock`, and some files were moved/renamed but weren't tracked properly. So I hope this is reasonably reviewable. Let me know if I can help make this easier!

@davidtheclark for review.

**Hold merge until...**
- [x] A new Batfish version has been published, and consumed in this PR. We need these two changes: https://github.com/mapbox/batfish/pull/119, https://github.com/mapbox/batfish/pull/117
- [x] Use the new `webpackStaticIgnore` Batfish configuration option for `expander.js` and `copy.js`
- [x] A solution for watching Assembly's CSS and JS has been implemented ([context](https://github.com/mapbox/assembly/pull/827#discussion_r129728845))


batfish.config.js
```diff
@@ -0,0 +1,40 @@
...
+      },
+      documentationData: () => {
+        return navigationStructure
+          .navigationList
+          .find((routeConfig) => routeConfig.name === 'Documentation' )
+          .props
+          .documentationData;
+      }
+    },
+    fileLoaderExtensions: (extensions) => extensions.concat(['svg', 'ico']),
```
davidtheclark: I think the Favicon should be added directly to the root of the pages directory. It should be copied from there to the same place in the built files.

batfish.config.js
```diff
@@ -0,0 +1,40 @@
...
+    notFoundPath: path.join(__dirname, './site/pages/404.js'),
+    pagesDirectory: path.join(__dirname, './site/pages'),
+    siteBasePath: '/assembly/',
+    webpackConfigStaticTransform: (webpackConfig) => {
+      // We need to ignore the following two files becasue their contents are executed immediately,
+      // and they rely on DOM APIs, so the static site renderer chokes on them otherwise.
+      webpackConfig.module.rules.unshift({
+        test: /copy\.js$|expander\.js$/,
+        loader: 'ignore-loader'
+      });
```
davidtheclark: Could you please open an issue about this in the Batfish repo?

package.json
```diff
@@ -9,15 +9,18 @@
...
     "build:css": "npm run generate-css && node scripts/build-css",
     "build:js": "node scripts/build-js",
     "build": "rm -rf dist && npm-run-all --parallel build:css build:js build:site",
     "lint:css": "stylelint src/*.css",
     "lint:js": "eslint .",
     "lint": "npm run lint:css && npm run lint:js",
-    "serve": "node scripts/serve",
-    "start": "npm run build:js && npm run serve",
+    "serve": "batfish serve-static",
+    "start": "npm run build:css && npm run build:js && npm run generate-site-data && batfish start",
```
davidtheclark: Without nodemon watching those CSS and JS files, what happens if you change one? Does the site rebuild?

package.json
```diff
@@ -36,11 +39,14 @@
   },
   "homepage": "https://github.com/mapbox/assembly#readme",
   "dependencies": {
+    "@mapbox/batfish": "^0.5.0",
     "autoprefixer": "^6.6.1",
     "concat-with-sourcemaps": "^1.0.4",
     "cp-file": "^4.1.1",
     "csso": "^3.1.1",
+    "es6-promise": "^4.1.1",
```
davidtheclark: This should be automatically loaded by Batfish: https://github.com/mapbox/batfish/blob/a50eea4fbd1b21f4f11bc7e63111e43769d6b9ad/lib/create-webpack-config-client.js#L22

site/navigation.js
```diff
@@ -1,10 +1,20 @@
...
+
 import { Logo } from './logo';
+import navigationData from '../_tmp_assembly/navigation.json';
 import pkg from '../package.json';
 
 class Navigation extends React.Component {
+  getActivePathname() {
+    // Remove the first section of the pathname (`/assembly/foo/` becomes `/foo/`).
+    return this.props.location.pathname.replace(/^\/[a-zA-Z0-9\-]*/, '');
+  }
```
davidtheclark, line 13: Maybe Batfish should expose an `unprefixUrl` function?

site/pages/404.js
```diff
@@ -0,0 +1,9 @@
+import React, { PureComponent } from 'react';
+
+export default class NotFound extends PureComponent {
+  render() {
+    return (
+      <div>Page not found.</div>
+    );
+  }
+}
```
davidtheclark: This page might not be necessary. If you hit a 404 on mapbox.com/assembly won't you land on the 404 page for mapbox.com?

davidtheclark (commented):
🥇 🥇 🥇 

A couple of things:

- This should wait on the next release of Batfish, which includes the fix for hash navigation.
- I notice that all the catalog sections are building out into individual pages. There should just be *one* page, right? That includes them all?

batfish.config.js
jfurrow: Oh, nice... done!

batfish.config.js
jfurrow: You're too fast, and already submitted a PR for the fix: https://github.com/mapbox/batfish/pull/119

I'd still be happy to create the issue if you want it to exist!

package.json
jfurrow: Good catch. Nope, with this change there's nothing watching the Assembly CSS and JS... I'll look into a proper solution to this.

package.json
jfurrow: Oh yeah, good catch, I meant to come back and debug this... removed & working as expected!

site/navigation.js
jfurrow: Oh, that would be nice!

site/pages/404.js
jfurrow: You're right, removed!

package.json
jfurrow: Would you mind reviewing my solution for this in [this commit](https://github.com/mapbox/assembly/pull/827/commits/30202b5de4ba074ed5cdc496795acb49ca35b03d)?

package.json
jfurrow: @davidtheclark Actually the build is still failing for me when I remove this from `package.json`... do you know what might be awry? Here's what I've been doing:
1. Manually remove `es6-promise` from `package.json`
2. `rm -rf node_modules`
3. `rm yarn.lock`
4. `yarn install`

That should work, right? When I do that, the build fails with ` Module not found: Error: Can't resolve 'es6-promise/auto'`

davidtheclark:
I published Batfish 0.6.0 with the fixes and features this branch needs.

Everything looks good! It's unfortunate that after `npm start` ctrl+c exits with a non-zero exit code, causing npm to complain —— but until we can find a fix it's not that big a deal.

Great work!

jfurrow:
@davidtheclark I updated Batfish to `0.6.0` and specified the `webpackStaticIgnore` option, so I think this is good to go.

Do you know why the Travis build is failing? There seems to be a bug with TravisCI preventing me from accessing this repo's logs.

davidtheclark (approved):
Someone (even you) can circle back to the kill problem later. For now, let's get this in.

jfurrow:
@davidtheclark I'd be happy to investigate the non-zero exit code business, especially since I introduced the ugliness 🙃

PR #829, opened 2017-07-28
Update changelog & prepare 0.16.0

jfurrow:
It doesn't seem like the Batfish migration belongs in the changelog, but I'll add if y'all think it's important.

@samanpwbb Review please?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
 # Changelog
 
+## 0.16.0
+
+- Add mobile conditions for `events-{none|all}`
+- Name change of `select-none` & `select-text` to `selectable` & `unselectable`
+- Add mobile conditions for `selectable` & `unselectable`
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
 # Changelog
 
+## 0.16.0
+
+- Add mobile conditions for `events-{none|all}`
+- Name change of `select-none` & `select-text` to `selectable` & `unselectable`
```
samanpwbb: can you add the [breaking] tag here?

samanpwbb (changes requested):
One little change, otherwise, looks good!

CHANGELOG.md
jfurrow: Good catch, done!

samanpwbb (approved):

PR #831, opened 2017-07-28
Specify npm version 5 for publisher

jfurrow:
Now that we're using npm version 5, publisher needs to know about it. https://github.com/mapbox/publisher#upgrading-npm

@davidtheclark Review?


davidtheclark (approved):
👍

PR #834, opened 2017-08-05
Update dependencies

davidtheclark:
The dependencies are falling a little behind here. I think it's worthwhile to bring this code up-to-date with the other stuff we're working on right now. All the changes in this PR should have no effect on the published files.


davidtheclark:
@jfurrow for review. I tried to keep the commit history really clean, one-step-at-a-time changes. Since each has a green checkmark next to it, each one represents a complete change. If there's anything you disagree with, we could then revert one of the commits.

Keep in mind that this involved updating Batfish after breaking changes, so double-checking the website build would be awesome.

jfurrow:
The changes look good! The build works mostly as expected, but the fonts are `404`ing:
screenshot 2017-08-07 10 09 05

davidtheclark:
Oh bummer. This has revealed to me a bug in the Batfish CSS system. Webpack is not copying assets that the CSS references (e.g. via `@font-face`).

davidtheclark:
@jfurrow This should be fixed now. Worked for me locally. Can you please verify?

jfurrow:
@davidtheclark The fonts are still `404`ing for me... am I doing something wrong?

It looks like I've got the updated `batfish`, right?
```
npm ls @mapbox/batfish

@mapbox/assembly@0.16.0 /Users/johnfurrow/Development/assembly
└── @mapbox/batfish@0.9.1
```

davidtheclark:
@jfurrow Using the latest Batfish release and undoing a couple of changes, I got the site running without font 404's in both development (start) and production (build & serve). If you can reproduce this, I think we're finally in the clear.

jfurrow (approved):
I'm happy to report the build is working flawless for me (both `npm run start` and `npm run build`). :shipit:

PR #840, opened 2017-08-09
Three things please

### LINKED ISSUE #808: Include built version in npm package
People with CSS build process will appreciate having the built version (the `dist/` directory) distributed with the npm package.

### LINKED ISSUE #838: Set <main> elements to display: block
IE11 does not support `<main>`: https://caniuse.com/#search=main

This does not frequently cause problems, or else we would have noticed before. But it *can* cause problems, because I just suffered from one of those problems this morning!

### LINKED ISSUE #839: scroll-styled should have momentum scrolling on iOS
To enable this, we need to add `-webkit-overflow-scrolling: touch;` to the rule.

davidtheclark:
- Momentum scrolling in iOS for scroll-styled …Closes #839.
- Reset main element to display block …Closes #838.
- Add dist to npm package …Closes #808.

@tristen for review.

After merging these I'd like to cut a new release.

### DIFF
--- a/src/miscellaneous.css
+++ b/src/miscellaneous.css
@@ -8,6 +8,8 @@
  * Style scrollbars. Only works in Webkit-based browsers (Safari and Chrome).
  * On dark backgrounds, add the `scroll-styled--dark` modifier.
  *
+ * **Use in combination with `scroll-auto`.**
+ *
  * @memberof Miscellaneous
    @selectors .scroll-styled
  * @example
@@ -233,6 +235,7 @@
  */
 .scroll-auto {
   overflow: auto !important;
+  -webkit-overflow-scrolling: touch;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/miscellaneous.css
```diff
@@ -14,6 +14,9 @@
  * <div class="scroll-styled scroll-auto h120"><div class='h240'></div></div>
  * <div class="scroll-styled scroll-styled--dark bg-gray-dark scroll-auto h120"><div class='h240'></div></div>
 */
+.scroll-styled {
```
tristen: This doesn't strictly need to fall under `scroll-styled` right? Should this be added to `scroll-auto` instead?

tristen (commented):
:one: 👏  :two: 👏 :three: 👏

src/miscellaneous.css
davidtheclark: I guess so, you're right.

davidtheclark:
Tried some real Git-trickery to amend that first commit. I think I got it.

tristen (approved):

PR #841, opened 2017-08-09
Remove border on disabled stroked buttons

### LINKED ISSUE #837: Remove stroke on disabled stroked button?
I'd like to propose a minor stylistic change to buttons. Currently, when a stroked button is disabled, the background grays out and the stroke fades:

screen shot 2017-08-08 at 12 58 39 pm

I think the combination of a background class and the stroke is a little bit off, I would prefer that stroked buttons appear like the standard disabled Assembly button (to maintain dimensions, we'd achieve this with colored strokes):

screen shot 2017-08-08 at 1 00 47 pm

Any objections or opinions on this?

dasulit:
screen shot 2017-08-09 at 12 07 00 pm

Closes #837. @davidtheclark for review

### DIFF
--- a/src/buttons.css
+++ b/src/buttons.css
@@ -129,7 +129,7 @@
   pointer-events: none;
   color: var(--inactive-text-color) !important;
   background-color: var(--disabled-primary-interactive-color) !important;
-  border-color: transparent !important;
+  box-shadow: none !important;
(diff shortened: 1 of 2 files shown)


davidtheclark (approved):
🚈

PR #845, opened 2017-08-18
Respond to switch hover state from its container

### LINKED ISSUE #844: Change `.switch:hover` to `.switch-container:hover > .switch`
That way, inline labels effects hover state of the input element like the checkbox and radio components do!

https://github.com/mapbox/assembly/blob/c0a91a1b748447aaa7aef4f9b3f92e5bbb3154f7/src/forms.css#L926-L929

tristen:
- Fixes #844

### DIFF
--- a/src/forms.css
+++ b/src/forms.css
@@ -924,7 +924,7 @@ input:checked + .checkbox {
 }
 
 /* state management for switches */
-.switch:hover {
+.switch-container:hover > .switch {
   color: var(--default-primary-interactive-color);
 }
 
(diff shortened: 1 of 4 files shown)


src/forms.css
```diff
@@ -924,7 +924,7 @@ input:checked + .checkbox {
 }
 
 /* state management for switches */
-.switch:hover {
```
samanpwbb, line 927: You'll need to update the switch rules in build-color-variants if you want this to work properly.

samanpwbb (changes requested):

src/forms.css
tristen: Hah, that's right 🤦‍♂️. Should be addressed now.

samanpwbb (approved):
🚜

PR #846, opened 2017-08-21
0.17.1

tristen:
Hmm there's a few changes I didn't capture in the CHANGELOG entry for 0.17.1. And it looks like this is including 0.17.0 🤔


samanpwbb (approved):
Looks great!

PR #852, opened 2017-09-11
make small range input smaller

### LINKED ISSUE #851: Make small range input smaller
We've been running into cases where we need a smaller range input. I think we should be find just shrinking the existing s range input a bit rather than adding a new size.

samanpwbb:
New smaller range input:

screen shot 2017-09-11 at 12 10 05 pm

closes #851

### DIFF
--- a/src/forms.css
+++ b/src/forms.css
@@ -517,24 +517,33 @@ because it seems to be more cross-browser consistent with all HTML5 input types 
 }
 
 /* range small track */
-.range--s > input::-ms-track { border-width: 9px 0; }
+.range--s > input::-webkit-slider-runnable-track {
+  height: 2px;
+}
+.range.range--s > input::-moz-range-track {
+  height: 2px;
+}
+.range.range--s > input::-ms-track {
+  border-width: 9px 0;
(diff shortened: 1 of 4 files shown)


samanpwbb:
Note: when building the repo, I noticed some mysteriously uncommitted changes to color-variants. These should have been committed, so I committed them here.

src/forms.css
```diff
@@ -517,24 +517,33 @@ because it seems to be more cross-browser consistent with all HTML5 input types
...
+  width: 12px;
+  height: 12px;
+  margin-top: -5px;
 }
 
 .range--s > input::-ms-thumb {
-  width: 16px;
-  height: 16px;
+  width: 8px;
+  height: 8px;
```
dasulit: These are smaller in edge, is that on purpose?

samanpwbb: Accident! Just fixed.

dasulit (approved):
Great! I like this new size, and ranges look good across browsers 👍  I don't know how to evaluate the color-variant diff, though, so... :woman_shrugging:

PR #865, opened 2017-10-14
disabled button should not use pointer cursor, closes #863

### LINKED ISSUE #863: Add button:disabled { cursor: default }
Because of this: https://github.com/mapbox/assembly/blob/dev-pages/src/reset.css#L74

samanpwbb:
@tristen for review


tristen (approved):
💌

Issue #860, opened 2017-09-23
Cannot override disabled button colors


davidtheclark:
The rule for disabled buttons looks like this:

```css
.btn:disabled {
  pointer-events: none;
  color: var(--inactive-text-color) !important;
  background-color: var(--disabled-primary-interactive-color) !important;
  box-shadow: none !important;
}
```

I'm not sure why those `!important`s would be necessary. And the effect they're having tonight is to block me from overriding the disabled button's background color with a `bg-*` class.

samanpwbb:
> I'm not sure why those !importants would be necessary.

Without `!important`, this disabled rule gets overridden by `btn--{color}` rules. So, I don't think this is a bug. We could adjust our strategy for styling disabled elements: switch to reducing opacity or something that else that doesn't overlap with btn modifiers, instead of setting a background color.

samanpwbb:
We could find some way for the following to be true:

```html
/* this is red */
<button class="btn btn--red">cat</button>

/* this is gray */
<button disabled class="btn btn--red">cat</button>

/* this is red */
<button disabled class="btn btn--red bg-red">cat</button>
```

But that seems a little weird to me...

davidtheclark:
> Without !important, this disabled rule gets overridden by btn--{color} rules.

It doesn't look like the `btn--{color}` classes use important: https://github.com/mapbox/assembly/blob/54d22b64b7b3f0580de2d60b977bb2964957b01a/scripts/build-color-variants.js#L148-L162. However, they do get beyond the specificity of a single class, because of `:hover` and `.is-active` — and I imagine that's what the disabled state's `!important` is meant to override. Seems we might be able to fix this by adding those selectors to the disabled rule?

samanpwbb:
> Seems we might be able to fix this by adding those selectors to the disabled rule?

Lets give this a try.

PR #866, opened 2017-10-14
Add styling for disabled elements with link class, closes #861

### LINKED ISSUE #861: Disabled links styled like disabled buttons
Disabled buttons have special styling. I think disabled `link`s should have the same, right?

samanpwbb:
@davidtheclark for review.

### DIFF
--- a/src/links.css
+++ b/src/links.css
@@ -35,3 +35,17 @@
 .link.is-active {
   color: var(--default-primary-interactive-color-dark);
 }
+
+/**
+ * When a form element like `button` includes both the `link` class and the `disabled` attribute, it will be styled accordingly.
+ * The disabled property has no effect on `a` elements.
+ *
+ * @memberof Links
+ * @example
+ * <button href='#Links' disabled class='link'>A disabled link</button>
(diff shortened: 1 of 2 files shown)


src/links.css
```diff
@@ -35,3 +35,17 @@
...
+ * The disabled property has no effect on `a` elements.
+ *
+ * @memberof Links
+ * @example
+ * <button href='#Links' disabled class='link'>A disabled link</button>
+ * <button href='#Links' disabled class='link link--red'>A disabled red link</button>
+ */
+.link:disabled {
+  cursor: default;
+  color: var(--inactive-text-color) !important;
```
davidtheclark, line 50: As mentioned in https://github.com/mapbox/assembly/issues/860#issuecomment-336714126, I wonder if there's a way to avoid `!important` here ...

samanpwbb: You okay with merging this PR, which follows the existing pattern, then make the fix in a bigger, new PR dedicated to purging assembly of !importants?

davidtheclark: Sure ✌️

PR #867, opened 2017-10-14
[breaking] remove importants and shorthand form border rules, closes #864

### LINKED ISSUE #864: Using media query variants with border classes results in unexpected colors
When using `border-l-mm`, this property is applied: `border-left: 1px solid !important;` Because this is using shorthand syntax, it implicitly sets the `border-left-color` value to `initial`, which can't overriden by a `border-color` class (specifically `border--gray-light` in my example). Both classes are using `!important`, but `border--gray-light` class presumably comes before the `border-l-mm` class in the source.

One solution might be to change the `border-l` family of classes to specifically apply the `border-width` and `border-style`, though this would increase the file size. Or we could alter the source order. Any thoughts on this?

The class list I'm using is: `border-t border-t--0-mm border-l-mm border--gray-light`

And here's the result:
screen shot 2017-10-03 at 11 18 49 am
screen shot 2017-10-03 at 11 18 39 am

samanpwbb:
@davidtheclark for review, cc: @jfurrow

### DIFF
--- a/src/theming.css
+++ b/src/theming.css
@@ -20,11 +20,26 @@
  * <div class='border'>border</div>
  * <div class='mt12 border-b'>border-b</div>
  */
-.border { border: 1px solid !important; }
-.border-t { border-top: 1px solid !important; }
-.border-r { border-right: 1px solid !important; }
-.border-b { border-bottom: 1px solid !important; }
-.border-l { border-left: 1px solid !important; }
+.border {
+  border-style: solid;
+  border-width: 1px;
+}
+.border-t {
+  border-style: solid;
+  border-top-width: 1px;
+}
+.border-r {
+  border-style: solid;
+  border-right-width: 1px;
+}
+.border-b {
+  border-style: solid;
+  border-bottom-width: 1px;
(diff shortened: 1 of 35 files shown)


samanpwbb (commented):
when border--2 is used now, it overrides directional border rules. Hm. I think we need to add specificity to solve this problem. Changes incoming.

samanpwbb:
Okay: I had to make breaking changes to border--2 but I think it's the best option.

src/theming.css
```diff
@@ -20,11 +20,26 @@
...
+  border-style: solid;
+  border-right-width: 1px;
+}
+.border-b {
+  border-style: solid;
+  border-bottom-width: 1px;
+}
+.border-l {
+  border-style: solid;
+  border-left: 1px solid;
```
davidtheclark: `border-left-width: 1px`

davidtheclark (approved):
🎖 one change to make before merging.

jfurrow:
🎉  Thanks for this, it looks great!

PR #869, opened 2017-10-15
[breaking] address a variety of form related issues

### LINKED ISSUE #855: Should input--s use txt-s size?
Default input size feels cramped. Here's the comparison:

input--s-size

### LINKED ISSUE #862: line height / txt-* class woes
There's an issue with "component" classes like form elements, button, icons. We've designed all these base component classes assuming assuming a default lineheight of 24px (which is the body default: https://github.com/mapbox/assembly/blob/54d22b64b7b3f0580de2d60b977bb2964957b01a/src/typography.css#L15).

When combining `.txt-{size}` classes with these elements, layouts don't end up looking right: button labels become un-centered, form elements take up the wrong amount of vertical space. The best way to solve the problem is to ensure all "component" style classes include a line-height rule that will override the `.txt-{size}` line height rule. We need to do take this approach because many component classes have built in top/bottom margin or padding that requires special line-height treatment to ensure alignment.

We're currently inconsistently applying line heights to form elements (see https://github.com/mapbox/assembly/blob/54d22b64b7b3f0580de2d60b977bb2964957b01a/src/buttons.css#L93 or https://github.com/mapbox/assembly/blob/54d22b64b7b3f0580de2d60b977bb2964957b01a/src/forms.css#L81) – we should do this everywhere.

I'm open to other approaches. We could go radically in the other direction and decouple line height from txt sizes and from form elements entirely. That would create problems of it's own though, and would require implementors to spend more time tweaking top/bottom margins and paddings of elements in order to ensure proper alignment (and would also require us to add more small-unit padding and margin classes like `px1`, `px2`, ect).

### LINKED ISSUE #868: Remove hover darken states from active switches/checkboxes/radios
It's weird that on hover, when a switch/checkbox/radio is active, that it takes on a dark color. There should be no style change on hover when these elements are active. This is a non-standard convention and we should drop it. Here's what I'm referring to:

samanpwbb:
This PR does a bunch of stuff:
- Improves styling of --xs and --s rules by default.
- Removes hover states from form elements, closes #868
- Rationalizes use of line height and font size in form elements, closes #862
- Includes line heights for all --s and --xs form modifiers, closes #855
- Changes input and select areas to use box shadow for borders, which aligns them to pixel grid -
without the need for special cased widths or line heights. Note that this does add some complexity to applying shadows and shadow-on-focus rules to line heights – but I think that trade-off is worth it.
- Adds --xs modifiers to all form elements
- Shrinks CSS to __182 kB/22.6 kB(gzipped)__ from 187 kB/23.1 kB (gzipped) 🎉 

These are obviously breaking changes, so we must proceed with caution! cc: @tristen @davidtheclark

### DIFF
--- a/src/forms.css
+++ b/src/forms.css
@@ -30,7 +30,8 @@
  */
 .input,
 .textarea {
-  border: 1px solid var(--gray-light);
+  box-shadow: inset 0 0 0 1px var(--gray-light);
+  padding: 6px 12px;
   border-radius: var(--border-radius);
   transition: background-color var(--transition),
     border-color var(--transition);
@@ -40,17 +41,12 @@
 
 .input:focus,
 .textarea:focus {
-  border-color: var(--default-primary-interactive-color);
+  box-shadow: inset 0 0 0 1px var(--default-primary-interactive-color);
 }
 
 .input::placeholder,
 .textarea::placeholder {
   color: var(--inactive-text-color);
-}
-
-/* Without this, IE always has a scrollbar */
-.textarea {
-  overflow: auto;
 }
 
 /* Remove bad padding and unique cancel button from IE and Safari inputs */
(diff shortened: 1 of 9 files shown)


src/forms.css
```diff
@@ -72,27 +68,36 @@
...
+}
+
+/**
+ * Make an input extra small.
+ *
+ * @memberof Inputs & textareas
+ * @example
+ * <input class='input input--xs' placeholder='extra small' />
+ */
+.input--xs {
```
davidtheclark, line 97: This is ridiculously small. We want this?

davidtheclark (commented):
I'm fine with the changes. I think @angel and @mgosline should take a look to make sure we're all on the same page.

src/forms.css
samanpwbb: We use similar inputs in Studio by add txt-xs and padding rules to `.input`. Would be cleaner to have access to this!

davidtheclark: Ok. Yipes.

tristen (approved):
Changes look good to me! 

- I think you are eluding to this but tabbing over inputs to trigger their :focus state no longer have the thicker light blue box shadow (because a focus border was replaced with box shadow). It's a little confusing but fine I guess!
- The "breaking changes" are visual ones right? otherwise there are just new additions?

samanpwbb:
> It's a little confusing but fine I guess!

This is my conclusion as well. The benefit of not having to account for the extra 2 pixels of vertical space exceeds the slightly weird styling you get on tab. It's no weirder than most every other css framework's handling of tab.

> The "breaking changes" are visual ones right?

Shadows and borders on form inputs would break pretty badly after this. input text sizes would also change, which could throw off layouts. We'll need to be deliberate about moving to whatever release of assembly includes these changes.

PR #870, opened 2017-10-15
[breaking] simplify and improve icon class implementation, closes #763

### LINKED ISSUE #763: Better strategy for vertically centering inline icons

samanpwbb:
screen shot 2017-10-14 at 7 17 23 pm

As long as icons are sized proportionally to text, they will not mess with text line height, and then it's easy to vertically center icons with 'icon inline-block align-middle'. If you need an icon at a specific size, use the h{size} and w{size} rules. Simple!

This commit removes icon-inliner, icon--s, and icon--l


davidtheclark:
This is an interesting idea. I do have some questions & concerns, though.

I think this does not end up making vertical centering icons with text much easier than the flexbox approach. Right now the catalog examples are *not* actually vertically centered with the text. 

Here's it looks like — the icon is too low:

screen shot 2017-10-15 at 7 21 36 am

Here's what it should look like, vertically centered:

screen shot 2017-10-15 at 7 21 41 am

In order for `inline-block align-middle` to work as you'd want, you have to (a) make sure all text you want vertically center with the icon is wrapped in elements, and (b) make sure that those text elements have `align-middle` applied. It's not enough to apply `align-middle` only to the icon. (The frequency with which I find myself repeating this trivia makes me expect that it's going to forgotten all over the place, and we'll end up misaligned icons.) 

So you end up having to apply just as many elements and about as many classes as if you'd used flexbox classes. The most notable difference is that you're using `vertical-align: middle` to center, which is easier to get wrong.

Also, this is a breaking change on a very widely used class. Probably we could mitigate it for our sites by making our `Icon` React component add size classes by default, and make this flexible-size behavior opt-in. Is that what you had in mind? Or are you thinking flexible-size should be the default? 

One consequence of flexible sizing is that we'll more often end up with icons that are pretty small. For example, on the site this

screen shot 2017-10-15 at 7 51 20 am

became this

screen shot 2017-10-15 at 7 51 26 am

samanpwbb:
> I think this does not end up making vertical centering icons with text much easier than the flexbox approach. Right now the catalog examples are not actually vertically centered with the text.

We can and should use flexbox when we don't need icons to flow with text. I don't think it's possible to make icons flow with text with flexbox.

> In order for inline-block align-middle to work as you'd want, you have to (a) make sure all text you want vertically center with the icon is wrapped in elements, and (b) make sure that those text elements have align-middle applied. It's not enough to apply align-middle only to the icon.

This is true. I have a new commit that addresses this problem by introducing two new classes.

> Also, this is a breaking change on a very widely used class. Probably we could mitigate it for our sites by making our Icon React component add size classes by default, and make this flexible-size behavior opt-in. Is that what you had in mind?

I had the reverse in mind. We'd default to context-based sizing, and make it easy to override. I think inheriting font size to set icon height will work well for us. We won't end up with awkwardly large or small icons as often.

src/icons.css
```diff
@@ -14,18 +14,53 @@
...
+ *   This is an icon
+ *   <svg class='icon icon--in-h'><use xlink:href='#icon-folder'/></svg>
+ *   inlined and centered with heading text.
+ * </h2>
+ */
+.icon--in-h {
+  display: inline-block;
+  vertical-align: top;
+  /* relies on expected txt to line height ratio of 1.2 */
+  height: calc(1em * 1.2);
```
davidtheclark: This could be `1.2em`, right?

src/icons.css
```diff
@@ -14,18 +14,53 @@
...
+ *   This is an icon
+ *   <svg class='icon icon--in-txt'><use xlink:href='#icon-folder'/></svg>
+ *   inlined and centered with body text.
+ * </p>
+ */
+.icon--in-txt {
+  display: inline-block;
+  vertical-align: top;
+  /* relies on expected txt to line height ratio of 1.5 */
+  height: calc(1em * 1.5);
```
davidtheclark: In regular paragraphs it is 1.6, though (15px/24px) — and the icon ends up too high:

screen shot 2017-10-15 at 4 16 10 pm

It's not too bad, at least not for this icon, but does illustrate the fragility of this approach.

davidtheclark (commented):
Ok, so now there are two distinct proposals, right? Changing the default icon size to `1em`, and introducing these new classes. It seems both could work independently?

I'm indifferent to the icon size question — I'm not the one whose consent is needed, I think.

How many situations do we have where a text is inline with icons? I'm not sure it's worth putting delicate rules into Assembly to cover these rare cases. Those edge cases could handle things on their own, I'm thinking.

davidtheclark:
I noticed that Google's material design icons punted on this problem, leaving it to each special case to figure out the method that works in its situation: https://github.com/google/material-design-icons/issues/121

src/icons.css
samanpwbb: Yes, I like the clarity of intent with calc (1em is the font size, multiply it by the line height). Either should be fine right?

src/icons.css
samanpwbb: This is hopefully close enough though –  much better than what we're using now.

samanpwbb:
I understand the criticisms, but I would be in favor of the changes proposed in this PR. Existing projects would benefit from these changes, and I would rather not require users to rediscover how to center align inline icons every time they use Assembly.

Happy to concede if the consensus is against me though. Added @tristen and @dasulit as reviewers.

src/icons.css
davidtheclark: Ok, fine by me.

tristen:
Just throwing it out there ... is another option to solve this upstream in mapbox-react-components? by adjusting the container size and viewbox of the svg element?

samanpwbb:
> is another option to solve this upstream in mapbox-react-components? by adjusting the container size and viewbox of the svg element?

Oh, it might be possible to solve it with javascript. We'd look at the element's computed line height value, then set the height of the element to match.

samanpwbb:
Decided to remove everything but the most basic `icon` class. Inline icon alignment should happen in JS by checking computed line height of the icon.

On the fence about whether the default icon size should use 1em x 1em (more convenient) or 18px x 18px (enforces icon pixel alignment by default). Either way, the expectation here is that icons would rely on `h` and `w` classes for sizing.

Would love another review from @davidtheclark @tristen when you two have a chance.

davidtheclark (approved):
🍈

PR #871, opened 2017-10-15
add no-print class, closes #842

### LINKED ISSUE #842: `no-print` class under @media print
Helpful to hide stuff like UI controls on pages that are designed to be printed (i.e. invoices)

samanpwbb:
@tristen for review


tristen (approved):
⛔️ 🖨

PR #874, opened 2017-10-23
remove importants from non-utility size classes and disabled classes

### LINKED ISSUE #860: Cannot override disabled button colors

### LINKED ISSUE #873: Audit use of `!important`
We're overusing `!important` right now. Important should only be used for utility classes that directly correspond to their value.

samanpwbb:
@davidtheclark for review. Does two things:

- Remove `!important` declarations from disabled component classes. Component modifiers _cannot_ override disabled state, but color / background utility classes (as well as inline styles) _can_ override disabled state. 
- Remove `!important` from size declarations on icons and loading classes.

Note that one contentious place where we continue to use `!important` after this PR is on triangles. Triangles should be `!important`, because the classes break in weird ways if any of the properties on the `triangle` class is overriden. 

closes #860, closes #873

### DIFF
--- a/src/buttons.css
+++ b/src/buttons.css
@@ -128,9 +128,20 @@
  */
 .btn:disabled {
   pointer-events: none;
-  color: var(--inactive-text-color) !important;
-  background-color: var(--disabled-primary-interactive-color) !important;
-  box-shadow: none !important;
+  color: var(--inactive-text-color);
+  background-color: var(--disabled-primary-interactive-color);
+  box-shadow: none;
+}
+
+.btn.btn--stroke:disabled {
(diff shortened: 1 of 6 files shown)


davidtheclark (approved):
Thumbs up!

PR #888, opened 2017-11-22
Vertically center left/right chevrons

jfurrow:
@angel noticed that the right chevron was slightly misaligned vertically. This PR fixes that for both right and left chevrons.

Here's a gif of me centering it in Illustrator:
chevron

@samanpwbb for review, please!


samanpwbb (approved):
Code looks good – but you'll need to update a snapshot test here!

jfurrow:
@samanpwbb Of course, should be fixed now!

Issue #893, opened 2017-12-14
font sizes in buttons and inputs


samanpwbb:
Right now, fonts in inputs and buttons all tend to be a little on the large side. We should either:
- Allow user to override form component's font declarations with txt-{size} font declarations
- Shrink some form and button font sizes ('regular' buttons use 's' font, small button use 'xs' font, ect).

Issue #892, opened 2017-12-14
--xs buttons and inputs should be 18px(?) tall instead of 15px tall


samanpwbb:
15px is too short for buttons and inputs.

PR #887, opened 2017-11-17
prepare assembly 1.0.0

samanpwbb:
Awaiting approval from @davidtheclark @tristen @dasulit. Will merge when I get three 👍 .

PR #894, opened 2017-12-14
Style cleanup

samanpwbb:
closes https://github.com/mapbox/assembly/issues/892, closes https://github.com/mapbox/assembly/issues/893, responds to @tristen's feedback in https://github.com/mapbox/assembly/pull/887, and drops unnecessary --xs modifiers and instead uses smaller styles for --s modifiers.

In general, this PR moves assembly to use styles more inline with how we style things in mapbox sites and apps.


src/forms.css
```diff
@@ -847,14 +798,38 @@ because it seems to be more cross-browser consistent with all HTML5 input types
   cursor: pointer;
   color: var(--default-secondary-interactive-color);
   font-weight: bold;
-  padding: 0 12px;
-  border-radius: 13px;
+  font-size: var(--font-size-s);
```
tristen, line 758: Missing a line-height rule to match this smaller size. It otherwise looks a little funky 

lineheight

tristen: For selects too the value is currently 24

screen shot 2017-12-14 at 10 22 09 am

tristen (changes requested):
😻

src/forms.css
samanpwbb: That's intentional: look at buttons on mapbox.com – i'm suggesting we use normal line height + small text for buttons.

davidtheclark: Look at buttons on Account — they have have larger text. Gotta coordinate the ideal here.

tristen: O_o

tristen (approved):
O_o

samanpwbb:
@davidtheclark @tristen for now, to get this PR through, what I can do is remove font size values in the standard button/toggle/select (so, users would need to add txt-s or txt-m to their buttons). We can then circle back around and find the perfect default button font size when we have a chance to convene with the design team.

davidtheclark:
I rebased this branch so we could revive its changes. 

I like the removal of `--xs` classes. They aren't used *widely*, so that should be fine.

I'm not sure I understand the line-height debate above. The reduction in font-size in buttons and button-like things definitely fits our usage, and the *heights* that we get in this PR match the vertical grid.

One thing I'm not so sure about is whether selects and toggle groups should match button styling. It seems to me that we always use them in places & ways where it makes more sense for them to match *input* styling: select looking like an input, toggle groups matching the height of adjacent inputs.

What do you think about making *that* change: selects and toggle groups would be as tall as inputs,  with the same font size as inputs. I might even go so far as to say that the default styling should be input influenced: gray borders and white background, instead of a button-like fill.

cc @mapbox/frontend-platform @dasulit @tristen @samanpwbb

samanpwbb:
> What do you think about making that change: selects and toggle groups would be as tall as inputs, with the same font size as inputs. I might even go so far as to say that the default styling should be input influenced: gray borders and white background, instead of a button-like fill.

More differentiation between buttons and other form elements sounds like a sensible direction to me, 👍.

davidtheclark:
This is all updated and ready for action. Will merge when there's a green light.

PR #897, opened 2018-01-27
Remove "Icons" from "Catalog" category

adnanh:
"Icons" are no longer nested under "Catalog"


davidtheclark:
Thanks, @adnanh! Sorry for the crazy delay merging this 😄

adnanh:
Better late than never! :-)))

PR #899, opened 2018-02-22
Fade in out animation

elifitch:
Added fade in out animation, suitable for a more subtle pulse effect.  Especially handy for loaders and skeleton screens.

### DIFF
--- a/src/animations.css
+++ b/src/animations.css
@@ -14,12 +14,14 @@
  * <div class="bg-blue inline-block mr18 w60 h60 round animation-spin animation-infinite"></div>
  * <div class="bg-blue inline-block mr18 w60 h60 round animation-fade-in animation--infinite"></div>
  * <div class="bg-blue inline-block mr18 w60 h60 round animation-fade-out animation--infinite"></div>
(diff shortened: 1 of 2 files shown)


src/animations.css
```diff
@@ -20,6 +20,7 @@
 .animation-spin { animation: spin 1.5s linear; }
 .animation-fade-in { animation: fadein 1.5s ease-in forwards; }
 .animation-fade-out { animation: fadein 1.5s ease-out reverse forwards; }
+.animation-fade-in-out { animation: fadeinout 1.5s linear forwards; }
```
samanpwbb: Can you add an example to the JSdoc comment so we get something to show up on the docs site?

samanpwbb (changes requested):
Add example for new rule docs then this is goo to go!

src/animations.css
elifitch: Jeepers, sure thing

elifitch:
@samanpwbb updated; changed the easing as well, I think this looks a bit less intense, more suitable I think

samanpwbb (approved):
Looks perfect 👍

PR #902, opened 2018-02-27
add suport for other subsets in opensans font

willemarcel:
A user opened an [issue on OSMCha repository](https://github.com/mapbox/osmcha-frontend/issues/270) alerting us that the OpenSans font we use doesn't support the Cyrillic alphabet. I've analysed it and confirmed this issue.

For example, see that text in the Google Fonts website:

screenshot-2018-2-27 google fonts

In [OSMCha](https://osmcha.mapbox.com/changesets/52287322) it is exhibited that way:
screenshot-2018-2-27 osm changeset analyzer
**Note the difference between the "a" in the first word and in "natural"**

So I converted the OpenSans font to woff and woff2 using https://transfonter.org/ with the option to maintain all the subsets. I tested in my machine using the new woff files and the result is:
screenshot-2018-2-27 osm changeset analyzer 1

This Pull Request update the woff files with the new I have generated.


samanpwbb:
Hi @willemarcel! I recommend using a custom build of Assembly and pulling in just the font you need (instructions for custom builds here: https://github.com/mapbox/assembly#custom-builds). This PR adds ~30KB to each font, which is a bit too heavy for most use-cases. We use a subset of Open Sans designed to be lean.

PR #903, opened 2018-03-27
Cleanup deps

davidtheclark:
I noticed that the npm package was installing some unnecessary dependencies, including react-helmet, which has peer dependencies so bugs me with a message. So this started when I looked into that.

I then decided to update everything, because it's been a while, Batfish has made a lot of improvements, etc.

There are very few source code changes that result from the updates:

- The Autoprefixer update ends up leaving out `-webkit-animation` vendor prefixes. https://github.com/mapbox/assembly/compare/cleanup-deps?expand=1#diff-90ba4c7a2e977423f2194db1decfe3c9R20401 These would only apply to some Android browsers around v4.4; and I think we should be ok trusting that it's a bug fix, not a regression, since Autoprefixer is so widely used.
- `postcss-custom-properties` changed its defaults, so I added options so it will behave the way we want: https://github.com/mapbox/assembly/compare/cleanup-deps?expand=1#diff-bdd03f78f49a7d785e978f33667932ddR103
- `svgo.optimize` now returns a Promise, so we could clean up some code around that: https://github.com/mapbox/assembly/compare/cleanup-deps?expand=1#diff-9866943e9f4ebdcf973ba09be6508396R53
- `mime` changed a function name: https://github.com/mapbox/assembly/compare/cleanup-deps?expand=1#diff-8dc589623193adc7d0f49a4c5b82bcd6R29
- Clipboard wants to be called `ClipboardJS` now: https://github.com/mapbox/assembly/compare/cleanup-deps?expand=1#diff-1ed10bb986fabd351e881baa10d5da83R1 I think this doesn't matter with the module import, but may as well be conventional.
- Prettier changed some spacing.

I've verified that the site builds well locally, both for development and production.

I think this would just cause a patch version release of Assembly. No API changes at all.

@samanpwbb for review.


samanpwbb (approved):
Nice! I pulled it down and was able to run everything without any problems.

PR #904, opened 2018-03-27
Add "from" option to a PostCSS invocation to avoid warning

davidtheclark:
With a very recent version of PostCSS, I started seeing this warning when running `buildUserAssets`:

> Without `from` option PostCSS could generate wrong source map and will not find Browserslist config. Set it to CSS file path or to `undefined` to prevent this warning.

With this small change, that warning goes away.

@tristen for review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.21.2
+
+- [fix] Avoid a PostCSS warning about `from` not being set.
+
 ## 0.21.1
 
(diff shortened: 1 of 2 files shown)


tristen (approved):
Strangely I'm not seeing this error npm installing and building from `master` but this branch builds just fine for me! 🐁

davidtheclark:
I only saw it when using `Assembly.buildUserAssets` from a different project. 🤷‍♂️ Harmless.

PR #908, opened 2018-06-26
Deploy fonts separately; reference them with absolute URLs

davidtheclark:
Font files do not need to be distributed with every version of Assembly. They should almost never change. So this PR tries to set up a new way of deploying and referencing fonts:

- Fonts are independently deployed to `api.mapbox.com/mapbox-assembly/fonts`.
- Fonts have `.v1.` inserted into their filename so that they can be cached forever, and if they ever *do* need to change (which [we have needed to do](https://github.com/mapbox/assembly/pull/786)) we don't have to worry about that cache.
- Fonts are referenced in the CSS with absolute URLs.
- Fonts are not distributed with the npm package.

After this change, the scripts `copy-fonts` and `copy-site-assets` are no longer needed.

@tristen this shouldn't work yet, because I need to deploy the fonts; but I was hoping for a quick glance from you first to make sure you agree with the approach. What do you think?


davidtheclark:
@tristen this is ready for another look. As discussed, I removed all the fonts from the repo. We'll deploy fonts as necessary with internal tooling.

We talked about allowing configuration of the font URL through a custom property. However, I learned that `var(--xxx)` is *not* evaluated within `url()` functions (because it might actually *be* a URL). So I think we can put that off for now. Since custom properties are becoming a thing that liberated people actually use sometimes, we might consider switching our system to https://github.com/postcss/postcss-simple-vars in order to avoid these kinds of headaches, doing the things we want to do without clashing with partially supported standards — since we'll need to be dealing with browsers that *don't* support custom properties for quite a while, probably.

tristen (approved):
Looks good & changes make sense to me!

PR #912, opened 2018-07-06
Drop custom build support for Node 4

davidtheclark:
I was looking into updating some dependencies and realized that many of the updates involve dropping support for Node 4. I don't think we have good reason to hang onto that, so let's drop it.

@danswick for review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## HEAD
+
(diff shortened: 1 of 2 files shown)


danswick (approved):
This seems like a very reasonable move to me, especially since it will only impact the build and development process. Just built and tested locally with node v6.14.3 and everything went smoothly. 👍

PR #913, opened 2018-07-06
Tidying and updates

davidtheclark:
Should follow #912, because many of these updates include dependencies dropping Node 6 support.

The most significant changes here are to ESLinting. I decided while updating ESLint we may as well incorporate eslint-config-mapbox, and that pointed out a few tweaks to make.

@danswick let's talk about these changes in person next week, because I think it would be worth pointing out some details about them to help you start to get familiar with the dependencies we often rely on.


davidtheclark:
Ping @danswick — maybe this fell off your radar?

danswick (approved):
Sorry about the delay, @davidtheclark! All of these changes look good to me. 

Just built and ran locally, clicked on the things that have been updated, and confirmed that everything seems to be working as expected! 

🍔 🐹 🔨

PR #917, opened 2018-08-03
Add switch--l for large switches

### LINKED ISSUE #882: Large switch component
There are some cases in which some folks have asked for a large switch component. Should we build a `switch--l` modifier?

davidtheclark:
Closes #882.

aug-02-2018 17-18-35

@angel for review, please.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## HEAD
 
 - [breaking] Drop support for Node 4 in custom build. **This has no effect on the client-side CSS or JS.**
+- [add] Add `switch--l`, for large switches.
 
 ## 0.22.0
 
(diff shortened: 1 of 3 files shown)


ahelkit:
@davidtheclark that doesn't look large :P I think the blue should be the Mapbox blue. Let me know if that is and the gif is just desaturated. Also, can the example have it with the on/off text. 
image
But vertically centered. 😂Let me know if that's not the feedback you were looking for or not! Thanks!

davidtheclark:
@angel The blue is Assembly's blue, so it will be Mapbox blue when using mapbox-assembly.

> that doesn't look large

I think you have pretty different standards of "large" from everyone else :) How large were you thinking, in pixels? The version above is 24px, so matching the line-height of regular-sized text.

ahelkit:
@davidtheclark Cool. Haha, 24px height works. I think it just looked small in the gif.  I played around with the size on the two-factor toggle in accounts and 24px looked good. Is small 18px?

PR #918, opened 2018-08-03
Style selects like inputs

davidtheclark:
Proposal!

Selects are used like inputs more than buttons. Often they are used right *next to* inputs, and their buttony styling makes them stand out much more than is desirable. Also: I can't find any public usage of the button-style select: seems that we always use `select--stroke` and then adjust some more to make it look kind of like an input.

So: what if we just make selects look like inputs, instead of buttons? That's what this PR does. Selects get the same border styling as inputs — the coloring also translates to their arrows.

Here's what we end up with:

screen shot 2018-08-02 at 7 44 02 pm

screen shot 2018-08-02 at 7 41 31 pm

I know @jseppi wanted something like this (I just can't find the issue).

What do the rest of you think, @mapbox/frontend?

@samanpwbb for review, please.

(Still need to add a changelog entry, if people are in agreement.)

### DIFF
--- a/src/forms.css
+++ b/src/forms.css
@@ -29,18 +29,20 @@
  * @memberof Forms
  */
 .input,
-.textarea {
+.textarea,
+.select {
   box-shadow: inset 0 0 0 1px var(--gray-light);
   padding: 6px 12px;
   border-radius: var(--border-radius);
   transition: background-color var(--transition),
-    border-color var(--transition);
+    box-shadow var(--transition);
   display: block;
   width: 100%;
 }
 
 .input:focus,
-.textarea:focus {
+.textarea:focus,
+.select:focus {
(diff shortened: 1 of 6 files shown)


src/forms.css
```diff
@@ -206,33 +209,27 @@ because it seems to be more cross-browser consistent with all HTML5 input types
...
-    background-color var(--transition);
-  border-radius: var(--border-radius);
-  background-color: var(--default-primary-interactive-color); /* match btn default state */
 }
 
 .select-arrow {
   position: absolute;
   right: 12px;
-  top: 50%;
+  top: calc(50% - 1px);
```
samanpwbb, line 228: 👌

samanpwbb (approved):
Looks great!

jseppi:
💯

PR #921, opened 2018-08-07
DO NOT MERGE!!!!!!

elfakyn:
cc @zmully @sidhe-elfakyn


elfakyn:
This is as should be. closing.

PR #928, opened 2018-12-04
Self-host placeholder image to avoid mixed content issues

danswick:
This PR adds a placeholder image, referenced in `typography.js`, to avoid the mixed content warnings triggered by the `placehold.it` placeholders originally used in the catalog page.


davidtheclark (approved):
Great 👍

PR #933, opened 2018-12-14
Fix a few details after switching domains

davidtheclark:
Each page should have its own `<link rel="canonical>` that points to *its* URL, not the homepage of the site. This should fix that, and also get us to deploy those tags to the new staging & production.

@kepta for review, please. cc @danswick as an FYI.

### DIFF
--- a/site/page.js
+++ b/site/page.js
@@ -2,7 +2,10 @@ import Helmet from 'react-helmet';
 import PropTypes from 'prop-types';
 import React from 'react';
 import { Navigation } from './navigation';
-import { prefixUrl } from '@mapbox/batfish/modules/prefix-url';
+import {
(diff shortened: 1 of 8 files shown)


site/page.js
```diff
@@ -34,7 +38,9 @@ class Page extends React.Component {
 }
 
 Page.propTypes = {
-  children: PropTypes.node.isRequired
+  children: PropTypes.node.isRequired,
+  // Should start and end with a slash.
```
davidtheclark, line 42: I thought a comment would be sufficient for this site, but for more high-pressure sites of course we'll want to automate any rules like this.

kepta (approved):
Great catch 🐌 !

PR #935, opened 2019-03-28
remove duplicate paths from history icon

samanpwbb:
Fixes an error in SVG. Allows the history icon to display correctly when used with opacity, and saves a few bytes.

@tristen for review


tristen (approved):
Looks great

PR #936, opened 2019-03-28
0.23.2

samanpwbb:
@kepta for review


danswick (approved):
tested the updated SVG and it looks good to me 🦁

PR #937, opened 2019-04-03
fix inter-doc links which pointed to localhost

andrewharvey:


davidtheclark (approved):
Thanks, @andrewharvey!

PR #946, opened 2020-10-10
Add position:sticky

tristen:


src/layout.css
```diff
@@ -224,6 +225,10 @@
 .absolute { position: absolute !important; }
 .relative { position: relative !important; }
 .static { position: static !important; }
+.sticky {
+  position: sticky !important;
+  position: -webkit-sticky !important; /* stylelint-disable-line value-no-vendor-prefix */
```
tristen: Safari still requires this. Unless there's a different way to handle adding this, I think disabling the style lint rule here makes sense.

samanpwbb: We use the autoprefixer postCSS plugin so there's no need for prefixes: https://github.com/mapbox/assembly/blob/publisher-staging/scripts/build-css.js#L114-L116

samanpwbb (approved):

samanpwbb (approved):
@tristen looks like you have a snapshot to update here.

katydecorah:
Does this class include media query sizes? If not, I think we should consider adding them. The docs team created `.sticky-mm` and `.sticky-mxl` to handle our sticky sidebars on larger screens.

tristen:
> Does this class include media query sizes? If not, I think we should consider adding them. The docs team created `.sticky-mm` and `.sticky-mxl` to handle our sticky sidebars on larger screens.

Ah! good catch. Updating to include those 👍

katydecorah (approved):
Perfect!

PR #947, opened 2020-10-10
Add alternative txt-strong rule

tristen:
I am constantly forgetting that a bold style is not `txt-strong` but `txt-bold`. I think we should include this as It's cheap to support both and [aligns with the HTML tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong#:~:text=The%20HTML%20Strong%20Importance%20Element,the%20contents%20in%20bold%20type.) in the same way that `txt-em` does.

### DIFF
--- a/src/typography.css
+++ b/src/typography.css
@@ -132,7 +132,8 @@ textarea {
  * @example
  * <div class='txt-bold'>txt-bold</div>
  */
-.txt-bold {
+.txt-bold,
+.txt-strong {
   font-weight: bold !important;
 }
 
--- a/test/__snapshots__/build-css.jest.js.snap
+++ b/test/__snapshots__/build-css.jest.js.snap
@@ -1003,7 +1003,8 @@ textarea{
   font-family:'Menlo', 'Bitstream Vera Sans Mono', 'Monaco', 'Consolas', monospace;
   font-size:90%;
 }
-.txt-bold{
+.txt-bold,
+.txt-strong{
   font-weight:bold !important;
 }
 .txt-h1{
@@ -12995,7 +12996,8 @@ textarea{
   font-family:'Menlo', 'Bitstream Vera Sans Mono', 'Monaco', 'Consolas', monospace;
   font-size:90%;
 }
-.txt-bold{
+.txt-bold,
+.txt-strong{
   font-weight:bold !important;
 }
 .txt-h1{


samanpwbb:
@tristen I don't think we should merge this. txt-bold conforms to the css `font-weight: bold`. It was a mistake to  use `txt-em` instead of `txt-italic`.

A better fix would be to add `txt-italic` and deprecate `txt-em`.

samanpwbb (changes requested):
for consistency, and to keep assembly's surface area small, we should add `txt-italic` and deprecate `txt-em`.

tristen:
> txt-bold conforms to the css

I wouldn't say that's a hard and fast rule for Assembly. We break that convention for most `txt-` classes (i.e they largely resemble HTML element names). 

> for consistency, and to keep assembly's surface area small

It's trivial to the size of Assembly to add an alternate `txt-strong` (or yeah `txt-italic`) I still think we should break consistency for ease.

samanpwbb:
> It's trivial to the size of Assembly

Size isn't the issue. Thinking more about user ergonomics: it's easier to understand and work with assembly code if we keep the ruleset small.

> I wouldn't say that's a hard and fast rule for Assembly. We break that convention for most `txt-` classes (i.e they largely resemble HTML attributes).

We're inconsistent, but I think experience has shown us that naming based on CSS styling is usually best and what we often expect in other parts of the library (https://github.com/mapbox/assembly/pull/948)

katydecorah:
I find the problem with having `txt-bold` and `txt-strong` is not understanding _why_ there are two classes that yield the same result. I think this could lead to some (minor) confusion on the difference and require us to maintain consistency in other codebases. 

For example, if I'm reviewing a PR where a contributor introduces `txt-strong` where the codebase already uses `txt-bold`, I would want to ask them use `txt-bold`, but at the same time I don't have a good reason why since the classes yield the same result. I prefer that Assembly enforces consistency so that we don't have to.

Overall, I have no preference on what we name the class that provides bold or italics, but prefer that there is only one class name to return bold or italics.

chrisbrainerd:
Agreed with @katydecorah, though I'd lean that we name it after the css rule, not the html tag (`txt-bold`). Keeping them all matching CSS rules makes it way easier to pick up Assembly if you already know CSS - I've run into lots of learning curve issues learning the way we name flexbox stuff since it doesn't match the actual css rules.

tristen:
👍 I'm going to close this. `txt-bold` ftw

PR #948, opened 2020-10-10
Rename scroll-* rules to overflow-*

### LINKED ISSUE #943: Rename `scroll-hidden` to `clip`
I think we should revert https://github.com/mapbox/assembly/issues/847. The argument was that it follows the `scroll-{value}` naming convention. In practice though, `overflow:hidden` isn't always used to target scroll behavior. One technique is to clip images in rounded containers. `clip` is a general enough name to capture both scenarios and is much easier to remember 😤

tristen:
Closes #943

### DIFF
--- a/site/catalog/bleeds.js
+++ b/site/catalog/bleeds.js
@@ -8,22 +8,22 @@ export class Bleeds extends React.Component {
           Bleeds
         </h2>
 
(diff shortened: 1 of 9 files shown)


samanpwbb (approved):
@tristen this is a good change. Can you target the `1.0.0` branch?

PR #949, opened 2020-10-10
Deprecate limiter

### LINKED ISSUE #942: Add screen size variation classes to limiter
Sometimes you want page content to be full-screen unless viewed on a larger viewport `-mxl` where a limiter should be in place. It would be nice to have ` limiter--mxl` class to do that.

tristen:
Closes #942 

Here's the stats after changes:

```
CSS: 176 kB (minified) => 22.2 kB (gzipped)
SVG: 50.7 kB => 15.9 kB (gzipped)
```

vs what's currently `1.0.0`

```
CSS: 172 kB (minified) => 21.6 kB (gzipped)
SVG: 50.7 kB => 15.9 kB (gzipped)
```


samanpwbb (changes requested):
Limiter already has different behavior at different viewports - you can see that the post-css script is breaking because of this. Besides, I don't think this is the correct approach to solving your problem. 

Could we add instead add % width classes that have media query modifiers and deprecate limiter? This would be simpler and more in line with Assembly's philosophy.

tristen:
> Could we add instead add % width classes that have media query modifiers and deprecate limiter? This would be simpler and more in line with Assembly's philosophy.

I like this idea! I wonder if we can rely on the existing `col--` classes instead. `col--10 mx-auto px0` is a near fit for limiter apart from a pixel-defined max-width. I wonder if we should add some larger `wmax` values instead?

samanpwbb:
> I wonder if we should add some larger `wmax` values instead?

Yes, lets do this

tristen:
This needs another pass but is otherwise ready for review.

test/__snapshots__/build-media-variants.jest.js.snap
```diff
@@ -9,50 +9,13 @@ exports[`buildMediaVariants defaults 1`] = `
 .col-mm {
   display: block;
   max-width: 100%;
-  flex-basis: 0;
```
tristen, line 12: We should talk about this change. This changes the behaviour of `.col` but excluding it means we can omit `flex:none` that were previously on `col--n` selectors

site/pages/examples/sidebar-apps.js
```diff
@@ -2,9 +2,9 @@ import React from 'react';
 import { HtmlExample } from '../../html_example';
 import { Page } from '../../page';
 
-const basicSidebar = `<div class='flex-parent viewport-full relative overflow-hidden'>
+const basicSidebar = `<div class='flex-parent h-viewport-full relative overflow-hidden'>
   <div class='flex-child w-full w240-ml absolute static-ml left bottom'>
-    <div class='flex-parent flex-parent--column viewport-third h-full hmax-full bg-white'>
+    <div class='flex-parent flex-parent--column h-viewport-1/3 h-full hmax-full bg-white'>
```
samanpwbb: This example seems broken - left sidebar should be taller?

Screen Shot 2021-04-13 at 1 02 26 PM

The markup for these layout examples is poor. We should cut them or rewrite them to be cleaner.

src/layout.css
```diff
@@ -4,40 +4,6 @@
...
- *  <div class='limiter border border--darken10 h24'>
- *   limiter
- *  </div>
- * </div>
- */
-.limiter {
-  padding-left: 24px;
-  padding-right: 24px;
-  width: 100%;
-}
```
samanpwbb, line 22: it'd be great to see an example that fully recreates all the details of limiter, across screen sizes, with existing classes.

samanpwbb (changes requested):
- There are still cases of `limiter` in the documentation site's CSS 
- The documentation site is now full bleed, leading to much too large text line lengths. Looks like we need to add some larger max width classes before we can actually deprecate limiter.
- Did you look into porting `grid` to css grid? This might allow us to deprecate the col class and all the CSS rules related to gutters that can currently lead to style conflicts with padding and margin classes.  As it stands, it looks like the only real purpose of the `col` class is to be used in the `grid--gut{x} > .col` classes. Fine to ticket this as a follow-up, but it'd be nice to get this into 1.0.0.
- Can you update the 'layout scales' documentation page to include all the new classes?

samanpwbb:
Broke out the grid refactor as a separate issue: https://github.com/mapbox/assembly/issues/994

site/page.js
```diff
@@ -26,10 +26,10 @@ class Page extends React.Component {
           />
           <link rel="canonical" href={prefixUrlAbsolute(this.props.path)} />
         </Helmet>
-        <div className="overflow-auto viewport-full-mm w180-mm fixed-mm top left flex-parent-mm flex-parent--stretch-cross-mm">
+        <div className="overflow-auto h-viewport-full-mm w180-mm fixed-mm top left flex-parent-mm flex-parent--stretch-cross-mm">
           <Navigation />
         </div>
-        <div className="ml180-mm limiter w-auto pl60-mm pr60-mm mb60 mt24">
+        <div className="ml180-mm w-full px24 w-11/12-mm wmax1200-mm px60-mm mb60 mt24">
```
samanpwbb: This is not working correctly. There's vertical scroll and a lot of the page is cut off. We could do something like this:

```suggestion
        <div className="ml180-mm px24 wmax1200-mm px60-mm mb60 mt24">
```

But then you're not getting a nice proportionally sized container.

samanpwbb (changes requested):
@tristen we need to make sure these new classes can be used to build nice limiters. I'm seeing some issues with the implementation on the docs site. Besides that, changes look good, but lets really make sure we can actually replace the limiter with new classes. We might need max with percentage classes :/

site/page.js
tristen: Whoops! Want to have another look? We _should_ be able to do this without introducing more stuff.

site/page.js
```diff
@@ -29,7 +29,7 @@ class Page extends React.Component {
         <div className="overflow-auto h-viewport-full-mm w180-mm fixed-mm top left flex-parent-mm flex-parent--stretch-cross-mm">
           <Navigation />
         </div>
-        <div className="ml180-mm w-full px24 w-11/12-mm wmax1200-mm px60-mm mb60 mt24">
+        <div className="ml180-mm mx-auto px24 w-full w-auto-mm w-5/6-ml w-11/12-mxl wmax1200-mm px60-mm mb60 mt24">
```
samanpwbb: This change still results in horizontal scroll on most screen sizes. I am pretty sure you either need to more heavily restructure things, or you need to add -wmax percentage classes.

samanpwbb (changes requested):
@tristen Still not quite there. Seeing overflow at both the xl and l breakpoints:

Screen Shot 2021-04-26 at 3 32 09 PM
:

Screen Shot 2021-04-26 at 3 32 18 PM

site/page.js
samanpwbb: Hm I was just looking at the implementation before this PR, and it looks like we don't use a proportionally scaled limiter there either: we force `width: auto` with `.w-auto` - otherwise main branch also scrolls!

tristen: oh! it looks like the previous markup was using `w-auto` which was overriding the width rules in `limiter`. Dropping those width rules and favoring `w-auto` should get us there no?

src/layout.css
```diff
@@ -66,46 +32,11 @@
 .col {
   display: block;
   max-width: 100%;
-  flex-basis: 0;
```
samanpwbb, line 69: I missed this change in earlier review. It sounds like this value makes it so you can't apply other types of widths. Does it do anything else? Seems like a necessary change based on your approach but are we losing anything?

samanpwbb (approved):
Looks good, needs a thorough changelog entry (make sure to document that flex-basis change!)

src/layout.css
tristen: I'm not seeing a good reason to keep it! In practice, we typically define a width to each `col` element so I don't think we're really losing anything here?

PR #970, opened 2021-03-29
Update browser support, closes #968

### LINKED ISSUE #968: Officially drop IE11 support
See https://github.com/mapbox/assembly/blob/85f5ec6a85874d54c7c3eeb784e42254d41394a8/scripts/build-css.js#L69-L77

Lets adjust the browserslist settings.

samanpwbb:
Update browser support, closes #968

@katydecorah would you mind giving this a look?


katydecorah (commented):
Looks great! You will also want to run `npx browserslist@latest --update-db` to update and remove duplications to the caniuse-lite version in package-lock.json (ref: https://github.com/browserslist/browserslist#browsers-data-updating).

samanpwbb:
> You will also want to run `npx browserslist@latest --update-db`

Good catch – Browserslist is a transitive dependency in our case  (via autoprefixer). I updated autoprefixer, which looks like it did clean up caniuse-lite in package.lock.

katydecorah:
@samanpwbb Were you able to get the tests to pass locally? I'm having trouble and I think it's related to the autoprefixer bump. (Travis is having an outage right now, which would explain why this PR didn't trigger Travis.)

samanpwbb:
@katydecorah looks like we have some breaking changes in our dependency tree to work out!

samanpwbb:
@katydecorah it looks like I'm running into issues with this plugin: https://github.com/postcss/postcss-custom-properties. I think I'm using the new API correctly but custom properties aren't compiling down to values at all. It could be because our browserslist settings now fully support CSS variables. 

What do you think about removing this plugin, and shipping assembly with variables?

test/__snapshots__/build-css.jest.js.snap
```diff
@@ -12642,6 +12500,7 @@ legend{
 }
 .range > input::-webkit-slider-thumb{
   box-sizing:border-box;
+  -webkit-transition:background 0.125s;
```
samanpwbb, line 12503: Weird 🤔 - maybe we need the prefix on psuedo elements?

katydecorah (approved):
It's nice to be able to drop a bunch of vendor prefixes!

PR #975, opened 2021-03-31
Add font-display:swap to @font-face declarations

### LINKED ISSUE #969: Consider using `font-display: swap`
Should we use `font-display: swap` in each `@font-face` declaration? 

This [Lighthouse performance diagnostic](https://web.dev/font-display/) suggests we could save 50-60 ms per font:

> image

`font-display` is not supported in IE, but that may not be an issue if we drop IE support #968.

samanpwbb:
Add font-display:swap to @font-face declarations. Here's a GIF showing the visual impact:

There's a little layout shifting but the upside is we get a quicker initial render.

@katydecorah for review

closes #969

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add `font-display:swap` to all @font-face declarations.
 - [breaking] Update browser compatibility targets, drop IE11 support.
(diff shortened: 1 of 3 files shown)


katydecorah (approved):
> There's a little layout shifting but the upside is we get a quicker initial render.

I think this is a fair trade-off! Thanks for implementing.

PR #977, opened 2021-04-01
Rework typographic scale, closes #962

### LINKED ISSUE #962: Adjust typographic scale
Right now assembly doesn't offer enough text contrast for dense UIs. We should add a new font size between txt-m and txt-s: txt-ms that is size 14 with line height of 21. Additionally, we should tweak txt-m to be 16px rather than 15px, which will normalize it's line height to 1.5 in addition to being a bit more legible. Here's the new scale:

samanpwbb:
Rework of our typographic system based on feedback from Documentation and Studio teams.

  | Selector | Old value | New value |
  | -------- | :-------: | --------: |
  | .txt-h1  | 45px/54px | 36px/45px |
  | .txt-h2  | 35px/42px | 30px/36px |
  | .txt-h3  | 30px/36px | 24px/30px |
  | .txt-h4  | 18px/24px | 20px/25px |
  | .txt-h5  | 15px/18px | 16px/20px |
  | .txt-xl  | 30px/45px | 30px/45px |
  | .txt-l   | 18px/30px | 20px/30px |
  | .txt-m   | 15px/24px | 16px/24px |
  | .txt-ms  |    N/A    | 14px/21px |
  | .txt-s   | 12px/18px | 12px/18px |
  | .txt-xs  | 10px/15px | 10px/15px |

Before:
Screen Shot 2021-04-01 at 10 44 31 AM
Screen Shot 2021-04-01 at 10 44 08 AM

After:
Screen Shot 2021-04-01 at 10 44 23 AM

Screen Shot 2021-04-01 at 10 43 56 AM

### DIFF
--- a/src/variables.json
+++ b/src/variables.json
@@ -75,18 +75,33 @@
 
   "focus-shadow": "0 0 0 3px rgba(137, 199, 216, 0.65)",
 
+  "font-size-h1": "36px",
+  "font-size-h2": "30px",
+  "font-size-h3": "24px",
+  "font-size-h4": "20px",
+  "font-size-h5": "16px",
+
+  "line-height-h1": "45px",
+  "line-height-h2": "36px",
+  "line-height-h3": "30px",
+  "line-height-h4": "25px",
+  "line-height-h5": "20px",
+
   "font-size-xl": "30px",
-  "font-size-l": "18px",
-  "font-size-m": "15px",
+  "font-size-l": "20px",
+  "font-size-m": "16px",
+  "font-size-ms": "14px",
   "font-size-s": "12px",
   "font-size-xs": "10px",
 
   "line-height-xl": "45px",
   "line-height-l": "30px",
   "line-height-m": "24px",
+  "line-height-ms": "21px",
   "line-height-s": "18px",
   "line-height-xs": "15px",
 
   "font-stack-base": "'Open Sans', sans-serif",
-  "font-stack-mono": "'Menlo', 'Bitstream Vera Sans Mono', 'Monaco', 'Consolas', monospace"
+  "font-stack-mono":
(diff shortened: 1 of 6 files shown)


src/variables.json
```diff
@@ -75,18 +75,33 @@
...
-  "font-size-m": "15px",
+  "font-size-l": "20px",
+  "font-size-m": "16px",
+  "font-size-ms": "14px",
   "font-size-s": "12px",
   "font-size-xs": "10px",
 
   "line-height-xl": "45px",
   "line-height-l": "30px",
   "line-height-m": "24px",
```
katydecorah, line 99: @samanpwbb Since `font-size-m` increased by a pixel, what do you think about increasing `line-height-m` to `25px`? I tested these changes in a docs site and noticed the line-height felt a little tight. 


`line-height: 24px` - current | `line-height: 25px` - proposed
:---|:---
Screen Shot 2021-04-01 at 3 31 05 PM | Screen Shot 2021-04-01 at 3 31 23 PM

samanpwbb: 16/24 is a 1.5 ratio, which is the recommended rule of thumb for body text. 15/24 is a 1.6 ratio, and I agree in some contexts it is more elegant. 

I am hesitant to raise the default line height to 25, since 25 is misaligned with Assembly's 6px baseline grid. As a result, certain layouts would be off by a few pixels unless heights on container elements get hard-coded.

Here are some reasonable ways forward:

1. Revert to 15/24 for txt-m
2. Use larger, more elegant line heights for text inside `prose`.
3. Add dedicated line height classes?
4. Stick with 16/24

katydecorah: That makes sense. Let's stick with what we have here, I'd much prefer `txt-m` be at 16px and stay aligned.

katydecorah (approved):
Looks good!

tristen (approved):

PR #978, opened 2021-04-05
Remove showall buttons

### LINKED ISSUE #474: Remove all the show all buttons
I think once https://github.com/mapbox/assembly/issues/462 is solved, it's possible remove all the show all buttons and allow users to search any available classes on the page right away.

samanpwbb:
The show/hide button toggle implementation was buggy. This PR removes the show/hide toggle, adjusts font size based on number of rules in a set, and adds a max height to sets.

Screen Shot 2021-04-05 at 10 30 38 AM

@tristen for review

closes #474


site/entry.js
```diff
@@ -59,29 +59,24 @@ class Entry extends React.Component {
       }
     }
 
+    const selectorElClasses = `mr3 pr3 px3 color-blue-dark round bg-blue-faint mb3 inline-block ${
+      selectors.length > 1 ? 'txt-s' : ''
+    } ${selectors.length > 20 ? 'txt-xs' : ''}`;
```
tristen: The jump in size feels pretty weird here. Why not just stick to `txt-s` as overflow provides scrolling?

samanpwbb:
@tristen removed the -xs logic, back to you again

tristen (approved):
🌮

PR #979, opened 2021-04-05
Add imperative focus API and improve documentation around assembly.js

### LINKED ISSUE #953: Add API to toggle focus control
Often I wish I could turn on focus hints even if user didn't press tab (for example, after some other user interaction where their focus state was important to highlight). Right now the only way to override the default behavior of focus control is with something like this: 

```js
document.documentElement &&
  document.documentElement.setAttribute('data-assembly-focus-control', 'visible');
```

Ideally, there's an assembly method I could call so I don't need to specify the element and attribute in my own code, like `assembly.setFocusMode(boolean)`.

samanpwbb:
Add imperative focus API and improve documentation around assembly.js

closes #953

@tristen for review


tristen (approved):
LGTM

Issue #963, opened 2021-02-24
Adjust form controls


samanpwbb:
We should make a variety of tweaks to form controls to consistify their visual style. Here are some mock ups that illustrate proposed changes:

<img width="69" alt="Screen Shot 2021-02-24 at 11 59 22 AM" src="https://user-images.githubusercontent.com/108094/109058422-d758ea80-7697-11eb-8146-3d2bdc5957b5.png">
<img width="105" alt="Screen Shot 2021-02-24 at 11 59 17 AM" src="https://user-images.githubusercontent.com/108094/109058426-d7f18100-7697-11eb-9a6a-4377f7802d07.png">
<img width="456" alt="Screen Shot 2021-02-24 at 11 59 14 AM" src="https://user-images.githubusercontent.com/108094/109058428-d7f18100-7697-11eb-98df-f50384507339.png">
<img width="291" alt="Screen Shot 2021-02-24 at 11 59 09 AM" src="https://user-images.githubusercontent.com/108094/109058429-d88a1780-7697-11eb-812f-285e50e321ba.png">

The changes mostly come down to switching to 1px strokes around all form elements, rather than using a blend of 1px and 2px strokes.

---

The following controls need to be updated:
- [ ] switch
- [ ] radio
- [ ] checkbox
- [ ] range (#980)
- [ ] select
- [ ] button

samanpwbb:
Should our range control be redesigned so the stem is easier to grab? Something closer to this style 🤔 :

<img width="192" alt="Screen Shot 2021-04-05 at 2 32 33 PM" src="https://user-images.githubusercontent.com/108094/113629842-d367a300-961b-11eb-93db-28a66129b923.png">

adrianababakanian:
Thanks for breaking out the range input rework into a separate ticket @samanpwbb -- to confirm, that should mean this ticket is just scoped to making the 1px border-radius changes, correct?

samanpwbb:
> to confirm, that should mean this ticket is just scoped to making the 1px border-radius changes, correct?

There are a few other changes we're going to want to make too, from design system:
- Make the stroked variant of toggle the default style, which will also let us make a stroked version of the toggle fit into our baseline grid.
- Make it easier to make border-less select elements.
- Update default colors to match Figma.

samanpwbb:
Pulled out the select separately, which means we can close this issue once https://github.com/mapbox/assembly/pull/981 merges

adrianababakanian:
[...]

PR #981, opened 2021-04-06
Update form element styles

samanpwbb:
This PR introduces the following changes:

- Refactor internal color variables so there is one set of variables for default button and link colors, and another set of variables for form controls.
- Adjust stroke widths for toggles, checkboxes, and radios from 2px to 1px.
- Change default form control colors from blue to gray for all form types.
- Increase inset around `switch--l` from 2px to 3px.
- Reduce height of `toggle` form element from 36px to 30px, and height of `toggle--s` from 24px to 18px.
- Remove bold text rule from toggle elements.
- Adjust the styling of disabled form controls to better differentiate active disabled and inactive disabled states.

refs #963

@katydecorah and @adrianababakanian for review 🙏


katydecorah (approved):
Love these updates!

adrianababakanian (approved):
Looks great!

PR #988, opened 2021-04-07
Drop tiled layout example

### LINKED ISSUE #764: tile-layout example needs to be reworked
In this example are a lot of violations of some important patterns. We should rework it.

I'll try to get to this soon, just don't have the time at this very moment so wanted to create an issue.

tristen:
I hate 2017 tristen

Closes #764


samanpwbb (approved):
Works for me. Feels weird deleting a page without a redirect but we're in some backwater labs site so it should be fine.

PR #989, opened 2021-04-07
refactor viewport classes

samanpwbb:
@tristen for review.

This PR also introduces some fixes to the build system that allow us to use the / character in classnames.


scripts/build-media-variants.js
```diff
@@ -76,11 +76,16 @@ const targetClassArray = [
...
-  'viewport-full',
+  'h-viewport-1\\/3',
+  'h-viewport-1\\/2',
+  'h-viewport-2\\/3',
+  'h-viewport-9\\/10',
+  'h-viewport-full',
+  'hmax-viewport-1\\/3',
+  'hmax-viewport-1\\/2',
+  'hmax-viewport-2\\/3',
+  'hmax-viewport-9\\/10',
```
tristen, line 87: Thoughts on keeping this `-almost`? `9/10` is a weird fraction and `almost` has some nice semantic meaning.

samanpwbb: `almost` never felt right to me. 9/10 accurately describes this class using similar patterns in place elsewhere in assembly.


Reasons I like 9/10:
- If we extend the scale to add more stops, 9/10 will always mean 9/10
- The classname tells me exactly what the class is doing. With `almost`, I must look at the css declaration to know exactly what the class does.

test/__snapshots__/build-css.jest.js.snap
```diff
@@ -1437,11 +1437,16 @@ textarea{
...
+.h-viewport-1\\\\/3{ height:33.3333vh !important; }
+.h-viewport-1\\\\/2{ height:50vh !important; }
+.h-viewport-2\\\\/3{ height:66.6666vh !important; }
+.h-viewport-9\\\\/10{ height:90vh !important; }
+.h-viewport-full{ height:100vh !important; }
+.hmax-viewport-1\\\\/3{ max-height:33.3333vh !important; }
+.hmax-viewport-1\\\\/2{ max-height:50vh !important; }
+.hmax-viewport-2\\\\/3{ max-height:66.6666vh !important; }
+.hmax-viewport-9\\\\/10{ max-height:90vh !important; }
+.hmax-viewport-full{ max-height:100vh !important; }
```
tristen, line 1449: Hmm the extra escaping characters looks wrong here

samanpwbb: @tristen think there are a few layers of escapes going on (manual escape in CSS, then another couple escapes applied by the snapshot code in Jest) but I am pretty sure this is working correctly.

scripts/build-media-variants.js
tristen: Unless all fractions are there and standard across our usage of fractions, `9/10` just feels forgettable and you are fitting it into a self-inflicted system.

> If we extend the scale to add more stops, 9/10 will always mean 9/10

I highly doubt we will extend the scale for `vh` units.

---

Another alternative: Can we just drop it? I did a quick audit of its usage and it can quite easily be dropped. It existed for splash images where you wanted a little reveal to indicate there was more below the fold. This seems like a fine candidate as an inline rule.

samanpwbb:
Closing: we will incorporate changes to viewport classes in https://github.com/mapbox/assembly/pull/949, cc: @tristen

PR #991, opened 2021-04-09
Add select--border-0 modifier for borderless select

### LINKED ISSUE #982: Make it easier to style borderless select elements
We want to make it easier to add select elements that don't have borders and extra left and right padding:

Screen Shot 2021-04-06 at 1 02 17 PM

Right now, to make borderless select elements that look good, some override classes need to be applied to get the spacing to look right. 

bordered and borderless selects should be managed with modifier classes.

adrianababakanian:
Closes https://github.com/mapbox/assembly/issues/982

This PR adds the `select--border-0` modifier class to support borderless select elements out-of-the-box. It removes the `select`'s default box-shadow and left padding.

@samanpwbb for review -- I'm curious if you think the padding right of the `select-arrow` should also be reduced or removed, in which case there would be some additional work to be done here.


adrianababakanian:
Oops, going to close this since I forgot to update the base branch to 1.0.0 when opening the PR.

PR #992, opened 2021-04-09
Add select--stroke modifier and change to default borderless select

adrianababakanian:
Closes https://github.com/mapbox/assembly/issues/982

This PR adds the ~`select--border-0` modifier class to support borderless select elements out-of-the-box. It removes the `select`'s default box-shadow and left padding.~ `select--stroke` modifier and updates select elements to be unstroked by default.

@samanpwbb for review -- I'm curious if you think the padding right of the `select-arrow` should also be reduced or removed, in which case there would be some additional work to be done here.

### DIFF
--- a/scripts/build-color-variants.js
+++ b/scripts/build-color-variants.js
@@ -190,6 +190,14 @@ function buildColorVariants(variables, config) {
       if (isNotAccessibleForForms(color)) return result;
       const darkerShade = getDarkerShade(color);
       return (result += stripIndent(`
+        .select--${color} {
+          color: var(--${color});
+        }
+
+        .select--${color}:focus {
+          color: var(--${darkerShade});
+        }
+
         .select--${color} + .select-arrow {
           border-top-color: var(--${color});
         }
--- a/scripts/build-color-variants.js
+++ b/scripts/build-color-variants.js
@@ -229,7 +229,7 @@ function buildColorVariants(variables, config) {
           box-shadow: inset 0 0 0 1px var(--${darkerShade});
         }
 
-        .select--stroke.select--${color}:focus {
+        [data-assembly-focus-control='visible'] .select--stroke.select--${color}:focus {
           box-shadow: inset 0 0 0 1px var(--${darkerShade}), var(--focus-shadow);
         }
       `));
--- a/scripts/build-color-variants.js
+++ b/scripts/build-color-variants.js
@@ -221,9 +221,12 @@ function buildColorVariants(variables, config) {
         }
 
         .textarea--border-${color}:focus,
-        .input--border-${color}:focus,
-        .select--stroke.select--${color}:focus {
+        .input--border-${color}:focus {
           box-shadow: inset 0 0 0 1px var(--${darkerShade});
+        }
+
+        .select--stroke.select--${color}:focus {
+          box-shadow: inset 0 0 0 1px var(--${darkerShade}), var(--focus-shadow);
         }
       `));
     }, '');
--- a/scripts/build-color-variants.js
+++ b/scripts/build-color-variants.js
@@ -190,11 +190,11 @@ function buildColorVariants(variables, config) {
       if (isNotAccessibleForForms(color)) return result;
       const darkerShade = getDarkerShade(color);
       return (result += stripIndent(`
-        .select--border-${color} + .select-arrow {
+        .select--${color} + .select-arrow {
           border-top-color: var(--${color});
         }
 
-        .select--border-${color}:focus + .select-arrow {
+        .select--${color}:focus + .select-arrow {
           border-top-color: var(--${darkerShade});
         }
       `));
@@ -208,13 +208,13 @@ function buildColorVariants(variables, config) {
       return (result += stripIndent(`
         .textarea--border-${color},
         .input--border-${color},
-        .select--border-${color} {
+        .select--stroke.select--${color} {
           box-shadow: inset 0 0 0 1px var(--${color});
         }
 
         .textarea--border-${color}:focus,
         .input--border-${color}:focus,
-        .select--border-${color}:focus {
+        .select--stroke.select--${color}:focus {
           box-shadow: inset 0 0 0 1px var(--${darkerShade});
         }
       `));
--- a/scripts/build-color-variants.js
+++ b/scripts/build-color-variants.js
@@ -198,6 +198,10 @@ function buildColorVariants(variables, config) {
           color: var(--${darkerShade});
         }
 
+        .select--${color}:hover {
+          color: var(--${darkerShade});
+        }
+
         .select--${color} + .select-arrow {
           border-top-color: var(--${color});
         }
--- a/src/forms.css
+++ b/src/forms.css
@@ -192,7 +192,7 @@ because it seems to be more cross-browser consistent with all HTML5 input types 
   font-size: var(--font-size-m);
   line-height: var(--line-height-m);
   color: currentColor;
-  padding: 6px 30px 6px 12px; /* plus arrow */
+  padding: 6px 30px 6px 12px; /* plus arrow */ 
   cursor: pointer;
 }
 
@@ -267,6 +267,30 @@ because it seems to be more cross-browser consistent with all HTML5 input types 
 
 .select--s + .select-arrow {
   right: 8px;
+}
+
+/**
+ * Make a select component borderless.
+ *
+ * @memberof Selects
+ * @example
+ * <div class='select-container'>
+ *   <select class='select select--border-0'>
+ *     <option>one</option>
+ *     <option>two</option>
+ *   </select>
+ *   <div class='select-arrow'></div>
+ * </div>
+ * <div class='select-container'>
+ *   <select class='select select--border-0'>
+ *     <option>one</option>
+ *     <option>two</option>
+ *   </select>
+ *   <div class='select-arrow'></div>
+ * </div>
+ */
+.select--border-0 {
+  box-shadow: none !important;
 }
 
 /**
--- a/src/forms.css
+++ b/src/forms.css
@@ -320,7 +320,7 @@ because it seems to be more cross-browser consistent with all HTML5 input types 
  *   <div class='select-arrow'></div>
  * </div>
  */
- .select:disabled {
+.select:disabled {
   pointer-events: none;
   color: var(--inactive-text-color);
 }
--- a/src/forms.css
+++ b/src/forms.css
@@ -192,7 +192,7 @@ because it seems to be more cross-browser consistent with all HTML5 input types 
   font-size: var(--font-size-m);
   line-height: var(--line-height-m);
   color: currentColor;
-  padding: 6px 30px 6px 12px; /* plus arrow */ 
+  padding: 6px 30px 6px 12px; /* plus arrow */
   cursor: pointer;
 }
 
@@ -281,16 +281,10 @@ because it seems to be more cross-browser consistent with all HTML5 input types 
  *   </select>
  *   <div class='select-arrow'></div>
  * </div>
- * <div class='select-container'>
- *   <select class='select select--border-0'>
- *     <option>one</option>
- *     <option>two</option>
- *   </select>
- *   <div class='select-arrow'></div>
- * </div>
  */
 .select--border-0 {
   box-shadow: none !important;
+  padding: 6px 30px 6px 0;
 }
 
 /**
--- a/src/forms.css
+++ b/src/forms.css
@@ -41,14 +41,13 @@
 }
 
 .input:focus,
-.textarea:focus,
-.select.select--stroke:focus {
+.textarea:focus {
   box-shadow: inset 0 0 0 1px var(--default-form-color);
 }
 
-.select:focus {
+/* .select:focus {
   box-shadow: none;
-}
+} */
 
 .input::placeholder,
 .textarea::placeholder {
--- a/src/forms.css
+++ b/src/forms.css
@@ -216,6 +216,10 @@ because it seems to be more cross-browser consistent with all HTML5 input types 
 
 /** @endgroup */
 
+.select:hover {
+  color: var(--default-form-color-dark);
+}
+
 .select:focus + .select-arrow {
   border-top-color: var(--default-form-color);
 }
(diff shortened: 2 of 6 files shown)


samanpwbb (commented):
@adrianababakanian this looks like a solid start! 

> I'm curious if you think the padding right of the `select-arrow` should also be reduced or removed

Looking to the figma files for guidance, I think we do want a small amount of right padding yes!

Screen Shot 2021-04-12 at 9 24 36 AM

Before I approve or reject this, I have a couple questions: 

1. Did you consider the trade-offs of making the _stroke_ a modifier and the unstroked style the new base style? That might be more in keeping with Assembly's additive styling model, but at the cost of a change that might be considered "breaking" (although that's up for debate). 

2. Did you consider styling disabled/active states differently for unstroked selects? I wonder, for example, if we should only apply background-color changes to stroked selects.

adrianababakanian:
Thanks for the feedback @samanpwbb !

> Looking to the figma files for guidance, I think we do want a small amount of right padding yes!-

👍 

> Did you consider the trade-offs of making the stroke a modifier and the unstroked style the new base style? 

I didn't consider this, though I do see how making the unstroked style the default would be more in keeping with the additive styling model that you mentioned. Should we move forward with switching the modifier to something like `select--stroke`, similar to `btn--stroke`?

> Did you consider styling disabled/active states differently for unstroked selects?

Great callout. I do think the active state makes sense to keep the same as the stroked select -- i.e., the arrow color becomes darker and font color remains the same.

Screen Shot 2021-04-13 at 9 52 09 AM

But, we might want to handle the disabled state differently, as it doesn't look great with the current state of this PR given the lack of left padding:

Screen Shot 2021-04-13 at 9 56 56 AM

I'm not seeing anything explicitly in the style guide for how to handle this case -- do you have any thoughts?

samanpwbb:
> Should we move forward with switching the modifier to something like `select--stroke`, similar to `btn--stroke`?

I am on board – @katydecorah would you be okay with this change? It may require some work to migrate for you.


> I'm not seeing anything explicitly in the style guide for how to handle this case -- do you have any thoughts?

Ya, maybe something closer to how we do disabled `link`? disabled-colored text and disabled-colored triangle, and leave background transparent?

katydecorah:
> would you be okay with this change?

👍 Yes, I think that pattern makes sense

adrianababakanian:
When modifying `select` to be borderless by default, I made a few modifications:

- Added a `select--stroke` modifier to add a border + additional padding to the select element.
- Changed the `select--border-${color}` modifiers to `select--${color}`. They now have the effect of coloring just the arrow when used with `select`, and both the arrow and border when used in conjunction with the `select--stroke` modifier.

Screen Shot 2021-04-13 at 5 00 34 PM

@samanpwbb I wonder if it would make sense for the `select--${color}` modifiers to also change the font color for both the bordered and borderless versions. Referencing Figma, it looks like one of the bordered selects has a different font + arrow color from its border, whereas the other uses the same color for all three elements. Should the former be the default coloring, and the latter the result if a user explicitly sets a color via a modifier?

Screen Shot 2021-04-13 at 5 11 47 PM

samanpwbb:
> I wonder if it would make sense for the `select--${color}` modifiers to also change the font color for both the bordered and borderless versions. 

For unstroked, definitely.

Not sure about stroked - we could go either way.. I know @katydecorah originally modelled these after text inputs, which do not change text color.

> Referencing Figma, it looks like one of the bordered selects has a different font + arrow color from its border, whereas the other uses the same color for all three elements.

What you're seeing is the difference between the active and inactive state. Active state should use `-dark` variants for everything, while inactive state should have a light border, and default colored text and triangle. 

Also, looking at your screenshot, I wonder if we should adjust the default select style to use a darker triangle?

adrianababakanian:
Thanks for the clarifications and continued reviews @samanpwbb ! I updated the default arrow to use the default form color rather than the lighter default form color, so now it's the same as the font color. The active state now also uses `-dark` variants for everything. 

Screen Shot 2021-04-14 at 9 44 46 AM
Screen Shot 2021-04-14 at 9 42 55 AM

Right now the colored stroked selects use an updated font color as well. I like how it creates consistency with how unstroked selects handle color variants, but at the same time I see the value in keeping them modeled after text inputs (which don't change colors). @katydecorah I'd appreciate any thoughts you might have on this!


Screen Shot 2021-04-14 at 9 42 29 AM

src/forms.css
```diff
@@ -42,10 +42,14 @@
 
 .input:focus,
 .textarea:focus,
-.select:focus {
+.select.select--stroke:focus {
```
samanpwbb: I don't think we need the `.select` part of this selector, can be .select--stroke:focus

katydecorah:
> Right now the colored stroked selects use an updated font color as well. I like how it creates consistency with how unstroked selects handle color variants

@adrianababakanian While I love the way it looks, we should make sure that the font color for text has an [accessible contrast ratio](https://webaim.org/resources/contrastchecker/). My vote is to optimize for legibility over consistency. I know the Assembly color palette will be updated in v1, so perhaps the font color will soon have improved color contrast that would allow us to successfully update the font color without losing legibility.

src/forms.css
```diff
@@ -42,10 +42,14 @@
...
   box-shadow: inset 0 0 0 1px var(--default-form-color);
 }
 
+.select:focus {
+  box-shadow: none;
+}
```
samanpwbb, line 50: I think we can cut this rule.

src/forms.css
```diff
@@ -126,11 +130,11 @@
  */
 .input:disabled,
 .textarea:disabled,
-.select:disabled {
+.select--stroke:disabled {
   pointer-events: none;
   color: var(--darken50);
   background-color: var(--disabled-interactive-color);
-  box-shadow: inset 0 0 0 1px var(--disabled-interactive-color-dark);
+  box-shadow: inset 0 0 0 1px var(--disabled-interactive-color-dark) !important;
```
samanpwbb, line 136: Is this `!important` required?

samanpwbb (commented):
@adrianababakanian this is coming together! Noticed a couple minor issues.

Another thing I noticed, that is also present on previous version: we don't have good `[data-assembly-focus-control='visible']` styling for selects. Is this something we could fix? Notice how other form types take on a prominent `:focus` style, but selects do not.

adrianababakanian:
> My vote is to optimize for legibility over consistency.

Really great point @katydecorah ! I think that makes sense. Optimizing for legibility, I see three potential paths forward:

1. Use the default font color for both stroked and unstroked selects regardless of the color modifier. This means that applying color variants to the default selects only affects the arrow color, which does look a little funny.
2. Use the `-medium` variants for each color from the new style guide as the text + arrow colors when color modifiers are applied (for both stroked + unstroked elements).
3. Use the `-medium` variants for each color from the new style guide as the text + arrow colors for _just_ the default select when color modifiers are applied. Also use the `-medium` color variant for the arrow in the stroked selects, but use the default form color for the text color.

src/forms.css
adrianababakanian: Without this rule, the default focus styling is applied (shown below). I didn't think this looked great due to the lack of left padding -- what do you think? 

Screen Shot 2021-04-14 at 2 19 40 PM

src/forms.css
adrianababakanian: This one is a bit interesting. When the `.select:focus { box-shadow: none }` rule is included, the `.select` part of the selector is necessary. Otherwise, a select element using a `select--stroke` modifier loses its border when focused. Here's a screenshot where the focus is currently on the first stroked element when `.select` is removed from the above selector. 

Screen Shot 2021-04-14 at 2 24 45 PM


Would be happy to hear ideas for better approaches! This also might change depending on how we decided to handle the `.select:focus` rule (i.e. the comment below).

src/forms.css
adrianababakanian: Without the `!important`, the disabled `select--stroke` elements retain their colored borders (see below). I wonder if there's a better way to achieve the desired effect without using `!important`? I thought using it might make sense given the context -- that is, we always want disabled `select--stroke`s to use the disabled color variable.

Screen Shot 2021-04-14 at 2 28 29 PM

samanpwbb: makes sense to keep !important

src/forms.css
samanpwbb: aha, that makes sense.

src/forms.css
samanpwbb: Aha, the more prominent tab style, even though it's ugly, might actually be the best move. Right now tab active state for selects is disorienting because it's so subtle.

Thinking about this more, what if we did this: write an even higher priority tab state selector that used `[data-assembly-focus-control='visible']`, and that applied _two box shadows_: one that is a single stroke, and a second box shadow that matches the tab active states we use elsewhere in assemby (for links, buttons, ect).

adrianababakanian: > Right now tab active state for selects is disorienting because it's so subtle.

That's true, just changing the text + arrow color to be slightly darker isn't noticeable unless you know to look for it. 

> what if we did this: write an even higher priority tab state selector that used [data-assembly-focus-control='visible'], and that applied two box shadows: one that is a single stroke, and a second box shadow that matches the tab active states we use elsewhere in assemby (for links, buttons, ect).

This makes a lot of sense to me! Maybe with one caveat -- for stroked buttons, the focus styling actually uses just one box-shadow and removes the thinner, colored box-shadow applied via the relevant color modifier. What do you think about using the same approach for selects?

Screen Shot 2021-04-14 at 2 57 23 PM
Screen Shot 2021-04-14 at 2 57 14 PM


This does break consistency with how colored textareas handle focus, though. This would go beyond the scope of this PR, but maybe the correct approach is to update both selects and _also_ textareas to have two box-shadows when focused? To help illustrate, the yellow textarea here has a tab active state, but it's very subtle.

Screen Shot 2021-04-14 at 2 55 08 PM

samanpwbb: @adrianababakanian ooh thanks for documenting this. We should ticket making the focus state for textareas and text inputs more prominent, maybe also with the double box-shadow technique?

adrianababakanian: I went ahead with adding a double box-shadow for focused stroked selects in this PR per your suggestion @samanpwbb , I'll go ahead and open a ticket for improving focus states for textareas and text inputs too 👍

samanpwbb (changes requested):
@adrianababakanian this is looking _really_ good. Almost ready to merge. I see two tiny issues:

Default stroked tab state doesn't show double border:
Screen Shot 2021-04-15 at 9 52 39 AM

Then, after seeing this work and thinking through your notes about differences between stroked/unstroked selects, and thinking through how we use selects throughout our systems now, here's my recommendation: 

- Model unstroked select styling after _links_: leave default and active coloring as you have it now, and _add_ hover states. This also more closely models how Studio now styles unstroked selects 
- Model stroked select styling after _text inputs_: use default text coloring always, and only apply color variants to border and triangle.

adrianababakanian:
Thanks @samanpwbb ! I _think_ I might have pushed up [`bbf9a30` (#992)](https://github.com/mapbox/assembly/pull/992/commits/bbf9a30658a739d4ba087d6210e4867642603cb7), which should resolve

> Default stroked tab state doesn't show double border

, while you had already started reviewing -- let me know if the most recent commit fixes that ☝️ issue! I noticed that in hindsight too.

I'll add those recommendations for unstroked/stroked select styling too.

src/forms.css
```diff
@@ -262,13 +266,43 @@ because it seems to be more cross-browser consistent with all HTML5 input types
...
+  padding: 3px 24px 3px 0; /* plus arrow */
 }
 
 .select--s + .select-arrow {
   right: 8px;
 }
 
+/**
+ * Modify a select component so that it has a border.
+ * Set the border (and arrow) color by adding a `select--{color}` modifier to the `select` element along with the `select--stroke` modifier (`*-dark` colors are not available).
```
samanpwbb: @adrianababakanian this documentation should clarify that the select--stroke modifier changes how the color modifier gets interpreted.

adrianababakanian:
@samanpwbb this should be ready for your final review!

> Default stroked tab state doesn't show double border

Per https://github.com/mapbox/assembly/pull/992#issuecomment-820592184, should be good on this point.

> Model unstroked select styling after links: leave default and active coloring as you have it now, and add hover states. This also more closely models how Studio now styles unstroked selects

I made these udpates, but kept the default box-shadow on focus. Links don't use box-shadows -- let me know if you think we should keep or remove them for unstroked selects.

> Model stroked select styling after text inputs: use default text coloring always, and only apply color variants to border and triangle.

✅ 

> this documentation should clarify that the select--stroke modifier changes how the color modifier gets interpreted.

This clarification is discussed in the docs for both the default select group and the `select--stroke` modifier. Open to feedback on restructuring docs to make this clearer!

scripts/build-color-variants.js
```diff
@@ -208,15 +220,18 @@ function buildColorVariants(variables, config) {
...
         .textarea--border-${color}:focus,
-        .input--border-${color}:focus,
-        .select--border-${color}:focus {
+        .input--border-${color}:focus {
           box-shadow: inset 0 0 0 1px var(--${darkerShade});
         }
+
+        .select--stroke.select--${color}:focus {
+          box-shadow: inset 0 0 0 1px var(--${darkerShade}), var(--focus-shadow);
+        }
```
samanpwbb, line 234: We only want to use the double-shadow when `[data-assembly-focus-control='visible']` selector is true, so this should be:

```css
[data-assembly-focus-control='visible'] .select--stroke.select--${color}:focus {
  box-shadow: inset 0 0 0 1px var(--${darkerShade}), var(--focus-shadow);
}
```

samanpwbb (changes requested):
Noticed one small detail, once that's taken care of this should be good to go.

samanpwbb (approved):
This is great! 

I had one other optional thought when playing around with this: should we remove the default border-radius on non-stroked selects? Just to keep the base style a bit more minimal? We'd then apply it to the stroked variant only. I don't think this change is necessary but if you agree with me, might be worth sneaking in.

adrianababakanian:
Thanks @samanpwbb !

>  should we remove the default border-radius on non-stroked selects?

Hm, my initial thought was yes, but since we're applying a box-shadow to unstroked selects on focus, I think it might be nice to keep the default border-radius. That way when a default select is focused, the overall effect would be more consistent with the style guide.

Issue #961, opened 2021-02-24
Adjust color scale


samanpwbb:
Assembly color scale has problems:
- some 'normal' colors are too low contrast against white and light backgrounds
- 'faint' and 'light' variants don't provide enough background shades for some apps. 'light' is too dark to use as a background shade in many cases.

Here's the proposal that adds two new shades to the color palette, a 'lighter', that provides another background shade, and a 'medium' that provides a high contrast shade that works on light backgrounds:

<img width="554" alt="Screen Shot 2021-02-24 at 11 52 10 AM" src="https://user-images.githubusercontent.com/108094/109057623-d1aed500-7696-11eb-9e18-55d89208f662.png">

These changes will be non-breaking, and will not effect form control generation in any significant way.

samanpwbb:
@aparlato one thing to keep in mind as you start this work: We want base Assembly's palette to use  different colors from what we use internally in mbx assembly. The screenshot above covers mbx assembly. We should base then new lighter and medium shades on assembly's existing palette.

tristen:
Can we scale back Assembly's default palette tremendously? Like 🪓 any purple, violet, yellow equivalents and just promote extending Assembly with BYOColors instead?

samanpwbb:
I do think we still a good palette of colors for background and text (we could _maybe_ cut a color or two here), but agree about cutting down form, link, button colors from the default build!

PR #997, opened 2021-04-14
961 Adjust color scale

aparlato:
@samanpwbb I decided to take the suggested approach of adjusting hue and then minor adjustments from the MBX Assembly palette from the ticket. This made sense to keep the contrast similar. 

I'm not sure this goes far enough however to differentiate itself from the mbx palette. Let me know if I should take another pass.

Note that the images here say `medium` but we changed it to `deep`.

### Updated regular assembly color palette
image

### Previous comparison of mbx and regular assembly color palettes:
Before update.
**left:** mbx assembly | **right:** regular assembly

### New comparison of mbx (from #961) and regular assembly color palettes
After update.
**left:** mbx assembly | **right:** regular assembly

#### Comparison of new colors next to each other:
image

[Link to Figma](https://www.figma.com/file/zlixGX0hJl6ujy6CDztdtx/Assembly-color-scale-adjust?node-id=0%3A1)


samanpwbb:
@aparlato as part of this PR can you share the before/after results from `npm run check-size`?

scripts/build-color-variants.js
```diff
@@ -8,43 +8,59 @@ const defaultVariables = require('../src/variables');
...
   'blue',
   'blue-light',
+  'blue-lighter',
   'blue-faint',
 
   'purple-dark',
+  'purple-deep',
   'purple',
   'purple-light',
+  'purple-lighter',
```
samanpwbb, line 63: I think we're going to want to exclude deep and lighter from all form color variant building (only make these new colors available through  `color-`, `bg-`, and `link-`).

aparlato: @samanpwbb Should this say `color-` `bg-` and `border-`? Link is currently included in forms based on the existing functions. I just updated to remove buttons as well, so we should see just these three.

Also, the function name `isNotAccessibleExceptButtons` is kind of confusing since if it returns true, we don't make a button variant. Should it be named like the other functions, maybe `isNotAccessibleForButtons` or am I misunderstanding?

aparlato:
**Before:**
CSS: 172 kB (minified) => 21.6 kB (gzipped)
SVG: 50.7 kB => 15.9 kB (gzipped)

**After:** 
CSS: 181 kB (minified) => 22.6 kB (gzipped)
SVG: 50.7 kB => 15.9 kB (gzipped)

scripts/build-color-variants.js
samanpwbb: You are right! Agree with everything you're saying. Some of this conditional code is confusing, definitely. Looks like you've decoded it and therefore you should feel good about renaming it!

src/color-variants.css
```diff
@@ -1579,57 +1659,77 @@ input:checked + .switch--dot-transparent::after {
  * @memberof Background colors
  * @example
  * <div class='grid'>
```
samanpwbb, line 1901: Consider adding txt-s to this container so their aren't so many line breaks in the example:

Screen Shot 2021-04-15 at 10 24 43 AM

---

Screen Shot 2021-04-15 at 10 24 37 AM

src/color-variants.css
```diff
@@ -1343,56 +1343,76 @@ input:checked + .switch--dot-transparent::after {
...
+ *   <div class='col col--2 color-darken50'>color-darken50</div>
+ *   <div class='col col--2 color-darken75'>color-darken75</div>
+ *   <div class='col col--2 color-lighten10'>color-lighten10</div>
+ *   <div class='col col--2 color-lighten25'>color-lighten25</div>
+ *   <div class='col col--2 color-lighten50'>color-lighten50</div>
+ *   <div class='col col--2 color-lighten75'>color-lighten75</div>
+ *   <div class='col col--2 color-white'>color-white</div>
+ *   <div class='col col--2 color-black'>color-black</div>
+ *   <div class='col col--2 color-transparent'>color-transparent</div>
+ *   <div class='col col--2 color-text'>.color-text</div>
```
samanpwbb: Not introduced by your pr but why does this one example start with a `.` 🤔

samanpwbb (approved):
This looks great! 

Consider updating some of the examples to use the new variants (here, we might want `deep` on `lighter`:

Screen Shot 2021-04-15 at 10 29 04 AM

Issue #889, opened 2017-12-05
Add active state to range inputs


tristen:
Default browser style has an active state style when the range is in use: https://codepen.io/elifitch/pen/GOeYLd?q=range&limit=all&type=type-pens

I wonder if we should tweak the focus state as well. right now it appears on the bar:

<img width="746" alt="screen shot 2017-12-05 at 10 29 38 am" src="https://user-images.githubusercontent.com/61150/33615152-54bda290-d9a7-11e7-8e3e-d06748b68437.png">

Ideally this would be on the element a user interacts with (the round slider)

PR #999, opened 2021-04-15
Rework range input

### LINKED ISSUE #980: Rework range input so track is easier to click
Breaking this out from https://github.com/mapbox/assembly/issues/963 as it's own issue. Current range input is impractical because clicking the thin range track requires a lot of precision. We could either stick with the same visual design but make the click area for the track larger, or update the whole design.

adrianababakanian:
This PR closes https://github.com/mapbox/assembly/issues/980 and https://github.com/mapbox/assembly/issues/889

- Updates cursor to a pointer when inside the range input's clickable area. This makes it clearer that the user does not need to click _exactly_ on the pixels of the range track to update the position of the range thumb.
- Update range thumb's border-radius from 2px to 1px for consistency with style guide.
- Apply focus state to range thumb instead of track ~(_note: this point is still in progress_)~

### DIFF
--- a/src/focus.css
+++ b/src/focus.css
@@ -16,22 +16,22 @@
 }
 
 /* Range focus */
-[data-assembly-focus-control='visible'] input[type=range] {
+[data-assembly-focus-control='visible'] .range input[type=range] {
   box-shadow: none;
 }
-[data-assembly-focus-control='visible'] input[type=range]:focus::-webkit-slider-thumb {
+[data-assembly-focus-control='visible'] .range input[type=range]:focus::-webkit-slider-thumb {
   box-shadow: var(--focus-shadow);
 }
-[data-assembly-focus-control='visible'] input[type=range]:focus::-ms-fill-upper {
+[data-assembly-focus-control='visible'] .range input[type=range]:focus::-ms-fill-upper {
   box-shadow: var(--focus-shadow);
 }
-[data-assembly-focus-control='visible'] input[type=range]:focus::-ms-fill-lower {
+[data-assembly-focus-control='visible'] .range input[type=range]:focus::-ms-fill-lower {
   box-shadow: var(--focus-shadow);
 }
-[data-assembly-focus-control='visible'] input[type=range]:focus::-ms-thumb {
+[data-assembly-focus-control='visible'] .range input[type=range]:focus::-ms-thumb {
   box-shadow: var(--focus-shadow);
 }
-[data-assembly-focus-control='visible'] input[type=range]:focus::-moz-range-thumb {
+[data-assembly-focus-control='visible'] .range input[type=range]:focus::-moz-range-thumb {
   box-shadow: var(--focus-shadow);
 }
 
--- a/src/focus.css
+++ b/src/focus.css
@@ -16,22 +16,22 @@
 }
 
 /* Range focus */
-[data-assembly-focus-control='visible'] .range {
+[data-assembly-focus-control='visible'] input[type=range] {
   box-shadow: none;
 }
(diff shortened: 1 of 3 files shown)


adrianababakanian:
👋 @samanpwbb I'm still trying to get 

> Apply focus state to range thumb instead of track

working, but just wanted to let you know that the other points are reviewable currently.

adrianababakanian:
@samanpwbb this PR is fully ready for review now!

src/focus.css
```diff
@@ -16,22 +16,22 @@
...
-[data-assembly-focus-control='visible'] .range:focus::-ms-fill-lower {
+[data-assembly-focus-control='visible'] input[type=range]:focus::-ms-fill-lower {
   box-shadow: var(--focus-shadow);
 }
-[data-assembly-focus-control='visible'] .range:focus::-ms-thumb {
+[data-assembly-focus-control='visible'] input[type=range]:focus::-ms-thumb {
   box-shadow: var(--focus-shadow);
 }
-[data-assembly-focus-control='visible'] .range:focus::-moz-range-thumb {
+[data-assembly-focus-control='visible'] input[type=range]:focus::-moz-range-thumb {
```
samanpwbb: This style rule needs a classname in it. As a principle, Assembly should leave things unstyled, unless they use Assembly classes. Imagine a case where an Assembly user wanted the behavior of the range input type, but needed to give it a custom style.

src/forms.css
```diff
@@ -328,6 +328,7 @@ because it seems to be more cross-browser consistent with all HTML5 input types
   margin: 0;
   border: 0;
   background: var(--transparent);
+  cursor: pointer;
```
samanpwbb, line 330: 👍 - this does make me wonder if the handle should have a _grabber_ cursor that transitions into grabbing on active?

samanpwbb (changes requested):
This looks like a good start @adrianababakanian! Noticing two issues:

1. Make sure our focus style rule includes an assembly classname in it so we don't accidentally style non-Assembly ranges.
2. The active click area seems to be nice and large in Firefox, but not in Chrome. Can we investigate ways to increase the active area in Chrome?

src/focus.css
adrianababakanian: Ah, makes perfect sense!

adrianababakanian:
Thanks @samanpwbb , made some updates based on your feedback:

> Make sure our focus style rule includes an assembly classname in it so we don't accidentally style non-Assembly ranges.

Fixed by adding the `.range` classname back to the relevant selector 👍 

> The active click area seems to be nice and large in Firefox, but not in Chrome.

Thanks for catching this! I spent some time trying to research any well-documented solutions for handling this difference between Firefox and webkit-based browsers, but wasn't able to find anything specific. I ended up adding vertical padding to the range input element and then explicitly removing this padding for Firefox (since the clickable area already works as expected there). 

I'd appreciate feedback on this approach -- it feels a bit roundabout. 

> this does make me wonder if the handle should have a grabber cursor that transitions into grabbing on active?

I tried this out -- let me know what you think!

test/__snapshots__/build-css.jest.js.snap
```diff
@@ -639,13 +645,34 @@ legend{
...
+.range > input::-webkit-slider-thumb:active{
+  cursor:grabbing;
+}
+
+.range > input::-moz-range-thumb:active{
+  cursor:grabbing;
+}
+
+.range > input::-ms-thumb:active{
+  cursor:grabbing;
```
adrianababakanian, line 662: For some reason, the cursor was not being updated when I used a comma-separated list of selectors -- I'm not sure why this is the case.

samanpwbb (approved):
Looks great🏅

PR #1002, opened 2021-04-27
Refactor flex classes, closes #996

### LINKED ISSUE #966: Refresh commonly used icons
Assembly’s icons have served us well, but we need to touch them up. Two big reasons for the refresh 1) most of them were designed for non-retina screens and can look unpolished on modern displays (sharp edges, thick strokes, ect), and 2) the icons don’t have consistent visual weight. Here are the icons we should target, because they're most heavily used:

- close
- lock/unlock
- eye/noeye
- mountain
- pencil
- plus 
- undo/redo
- search
- share
- sprocket
- sun
- trash

@katydecorah do you have any requests from docs?

### LINKED ISSUE #996: Future of the flex-child rule?
Flex child has the following rules:

```
.flex-child {
  display: block;
  max-width: 100%;
}
```

This rule exists to handle IE11 bugs that we don't need to worry about any more (see https://github.com/philipwalton/flexbugs). I think we could deprecate this class. It would have the effect of really simplifying a lot of markup. Mapbox CSS tends to be packed full of `flex-child` declarations.

How should we handle flex-child modifiers? We could rename them to use a simpler form: `flex-grow` and `flex-shrink`. 

This would be a big change, so would love input from @katydecorah / @tristen / @dereklieu

samanpwbb:
This PR simplifies our flex classes. Remove `flex-child`, rename `flex-parent` to `flex`, rename `flex-parent-inline` to `inline-flex` (to match display declaration), rename all `flex-parent--` modifiers to `flex--`, and rename `flex-child--` modifiers to non-modifiers (`flex-child--grow` is now `flex-child-grow`).

@katydecorah for review - I made some changes that are different from your suggestions in the issue. This PR uses parent modifiers, and distinguishes child classes from modifiers and from flex display classes by prefixing them with flex-child. I found this approach to fit better with Assembly's naming conventions.


site/pages/examples/progress-bars.js
```diff
@@ -6,14 +6,14 @@ const basic = `<div class='h12 bg-darken10 relative round-full'>
...
+const playback = `<div class='flex flex--row flex--center-cross border border--gray-light round-l'>
+    <button class='py18 pr12 pl18 border-r border--gray-light bg-white bg-green-light-on-hover cursor-pointer round-l'>
       <div class='triangle--r triangle color-green color-green'></div>
     </button>
-    <div class='flex-child flex-child--grow px18 py18 bg-white'>
+    <div class='flex-child-grow px18 py18 bg-white'>
       <div class='relative w-full h12 bg-darken10 round-full'>
         <div class='absolute top right bottom left m3 bg-green-light round-full' style='width:50%;'>
-          <span class='absolute top right w30 h6 bg-green round-full'></span>
+          <span class='absolute top right w30 h-full bg-green round-full'></span>
```
katydecorah, line 16: It looks so much better 😄

site/pages/examples/tooltips.js
```diff
@@ -2,29 +2,29 @@ import React from 'react';
...
 </div>`;
 
-const basic_with_closure = `<div class='flex-parent-inline flex-parent--center-cross flex-parent--column'>
-  <div class='flex-child relative px18 py18 round txt-bold txt-s align-center bg-darken75 color-white'>
-    <button class='flex-child absolute top right'>
-      <svg class='flex-child icon link color-white'><use xlink:href='#icon-close'></use></svg>
+const basic_with_closure = `<div class='inline-flex flex--center-cross flex--column'>
+  <div class='relative px18 py18 round txt-bold txt-s align-center bg-darken75 color-white'>
+    <button class='absolute top right'>
+      <svg class='icon link color-white'><use xlink:href='#icon-close'></use></svg>
```
katydecorah: This is slightly out of scope for this PR, but the close icon is slightly larger than before and is tight against the tooltip. Here's a suggestion if we feel this is a good opportunity to fix:

```suggestion
      <svg class='icon link color-white mt3 mr3'><use xlink:href='#icon-close'></use></svg>
```

src/layout.css
```diff
@@ -335,205 +335,189 @@
...
-/* Specifically, the above addresses #1, #2, and #12 in https://github.com/philipwalton/flexbugs */
+.flex--space-between-main { justify-content: space-between !important; }
 
 /**
  * Make a child to grow to fill whatever space is available in the main axis of the parent container.
  * This is useful when you have one or more elements of
  * fixed width, and another element that should take up the remaining space in the row.
  *
- * Typically, you will also want to apply the `flex-child--no-shrink` modifier to the element in the row with a fixed width.
+ * Typically, you will also want to apply the `flex-child-no-shrink` to the element in the row with a fixed width.
```
katydecorah:
```suggestion
 * Typically, you will also want to apply `flex-child-no-shrink` to the element in the row with a fixed width.
```

or

```suggestion
 * Typically, you will also want to apply the `flex-child-no-shrink` class to the element in the row with a fixed width.
```

katydecorah (approved):
@samanpwbb I think your decision to keep the `flex-child-` prefix makes a lot of sense. I love how much cleaner these classes look. I left two small suggestions.

PR #1003, opened 2021-04-27
879 toc

samanpwbb:
Simple change that adds a 'home' entry and subnav to make it easier to find important docs that are on the home page: 

Screen Shot 2021-04-26 at 6 42 02 PM

@tristen for review


tristen (approved):
💯

PR #1012, opened 2021-05-17
Simplify color system

samanpwbb:
This PR fixes an upstream issue in Mapbox's build of Assembly but should also generally simplify Assembly for everyone. Not a breaking change but does result in subtle visual differences (generally I'd consider it an improvement - nicer disabled and semitransparent colors). Remove all colors specific to forms, buttons, links, use only generic colors. Add new disabled color variants.

@tristen for review

### DIFF
--- a/src/variables.json
+++ b/src/variables.json
@@ -1,5 +1,5 @@
 {
-  "gray-dark": "#2C3031",
+  "gray-dark": "#1a2224",
   "gray-deep": "#4D5255",
   "gray": "#757D82",
   "gray-light": "#C7CACC",
@@ -55,11 +55,11 @@
   "purple-lighter": "#E7E0F1",
   "purple-faint": "#F3F1F6",
 
-  "darken5": "rgba(0, 0, 0, 0.05)",
-  "darken10": "rgba(0, 0, 0, 0.1)",
-  "darken25": "rgba(0, 0, 0, 0.25)",
-  "darken50": "rgba(0, 0, 0, 0.5)",
-  "darken75": "rgba(0, 0, 0, 0.75)",
+  "darken5": "rgba(26, 34, 36, 0.05)",
+  "darken10": "rgba(26, 34, 36, 0.1)",
+  "darken25": "rgba(26, 34, 36, 0.25)",
+  "darken50": "rgba(26, 34, 36, 0.5)",
+  "darken75": "rgba(26, 34, 36, 0.75)",
 
   "lighten5": "rgba(255, 255, 255, 0.05)",
   "lighten10": "rgba(255, 255, 255, 0.1)",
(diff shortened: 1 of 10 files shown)


src/variables.json
```diff
@@ -71,19 +71,9 @@
...
-
-  "default-button-color": "#448ee4",
-  "default-button-color-dark": "#346db0",
-
-  "default-form-color-light": "#ccc",
-  "default-form-color": "#666",
-  "default-form-color-dark": "#2d2d2d",
+  "disabled10": "rgba(158, 163, 167, 0.1)",
+  "disabled25": "rgba(158, 163, 167, 0.25)",
+  "disabled75": "rgba(158, 163, 167, 0.75)",
```
samanpwbb, line 76: This color is exactly 50% between gray and gray light. It was picked because it's legible against both dark and light backgrounds unlike any other gray colors in the palette.

src/variables.json
```diff
@@ -71,19 +71,9 @@
...
-  "inactive-text-color": "rgba(64, 64, 64, 0.45)",
-
-  "text-color": "rgba(0, 0, 0, 0.75)",
-
-  "default-button-color": "#448ee4",
-  "default-button-color-dark": "#346db0",
-
-  "default-form-color-light": "#ccc",
-  "default-form-color": "#666",
-  "default-form-color-dark": "#2d2d2d",
```
samanpwbb, line 86: All instances of all these colors have been replaced by our generic colors.

src/variables.json
```diff
@@ -55,11 +55,11 @@
...
-  "darken5": "rgba(0, 0, 0, 0.05)",
-  "darken10": "rgba(0, 0, 0, 0.1)",
-  "darken25": "rgba(0, 0, 0, 0.25)",
-  "darken50": "rgba(0, 0, 0, 0.5)",
-  "darken75": "rgba(0, 0, 0, 0.75)",
+  "darken5": "rgba(26, 34, 36, 0.05)",
+  "darken10": "rgba(26, 34, 36, 0.1)",
+  "darken25": "rgba(26, 34, 36, 0.25)",
+  "darken50": "rgba(26, 34, 36, 0.5)",
+  "darken75": "rgba(26, 34, 36, 0.75)",
```
samanpwbb, line 62: the darken classes are all now based on gray-dark.

samanpwbb: 🤔 should we rgb everywhere instead of hex? annoying to translate between the two.

samanpwbb:
Hm looking at assembly now: I think we need to adjust active colors everywhere to be `-deep` rather than `-dark`.

src/variables.json
tristen: I kind of wish we would change all of these to HSL instead. It's so much more readable

tristen (approved):
Approving to unblock but can we table the custom disable work from this PR? I think increasing the color variable output with these gray variations against light and dark values needs more discussion.

Screen Shot 2021-05-18 at 11 44 20 AM

PR #1015, opened 2021-05-19
Add missing dash to example on Grid Catalog page

katydecorah:
This PR adds a missing dash to an example on the Grid Catalog page: https://labs.mapbox.com/assembly/catalog/#Grids where `w-1/2` was not being set:

Stage | Image
---|---
Current | image
Proposed | image


tristen (approved):
eeeek thx!!

PR #1017, opened 2021-05-20
Remove `flex-grow` from `col`

### LINKED ISSUE #1016: Remove `flex-grow` from `.col`
When using `grid` to build a `col` layout, if there are uneven number of items the last item will go full-width:

image
[Reduce test case](https://codepen.io/katydecorah/pen/4640926a8251b147692ec4d3374bd467?editors=1100)

This happens because `.col` sets `flex-grow: 1`. Since we have two options to allow items to grow `.col--auto` and `.flex-child-grow`, we should remove `flex-grow` from `.col`. 

With this change, it will change the concept of `.col` slightly since the class will depend on other to give its shape within the grid.

## Which classes benefit from flex-grow?

If we make this change, we lose the concept of the uncontrolled grid with the `.col` class. This means you will need to use a `.w-*` or `.col--auto` class to take advantage of the grid:

`.col` with `flex-grow` | `.col` without `flex-grow`
:---|:---
Screen Shot 2021-05-19 at 1 53 18 PM | Screen Shot 2021-05-19 at 1 53 22 PM

If we make this change, the %-based margins may need a `.w-*` or `.col--auto` class:

`.col` with `flex-grow` | `.col` without `flex-grow`
:---|:---
Screen Shot 2021-05-19 at 1 44 36 PM | Screen Shot 2021-05-19 at 1 44 10 PM

## What alternatives have you tried?

- Introduce a new class `flex-child-no-grow` that will disable `flex-grow`. This works, but adds code to Assembly.
- Use a right % margin on the last element. This solution works, but is a bit clunky and may not scale well.
- Removing `col` class from each element works, but then you cannot use `grid--gut*` classes to add spacing between elements.

katydecorah:
This PR introduces a breaking change to remove `flex-grow` from `.col`.

- Updates documentation and example for grid/col
- Updates documentation and example for %-based margin classes
- Removes "Uncontrolled grid" section from catalog

Fixes #1016


src/layout.css
```diff
@@ -12,17 +12,17 @@
  */
 
 /**
- * Grids require at minimum two elements: A parent with the class `grid` and a child with the class `col`. By default, each `col` in a grid will occupy equal space. The `grid` class includes `*-mm`, `*-ml`, and `*-mxl` variations to target screen sizes.
+ * Grids require at minimum two elements: A parent with the class `grid` and a child with the class `col`. To define the width of `col` in a grid, add a width class, `col--auto`, or both. The `grid` class includes `*-mm`, `*-ml`, and `*-mxl` variations to target screen sizes.
  *
- * When using these classes, do not apply left or right margin or padding directly to the \`grid\` or \`col\` elements or you will face unexpected side-effects.
+ * When using these classes, do not apply left or right margin or padding directly to the `grid` or `col` elements or you will face unexpected side-effects.
```
katydecorah: Is this statement still true? The %-based margin examples in `scripts/build-layout-scales.js` use a `col` and margin class. Should this PR update the margin class example so that we are not applying the margin on the `col`?

samanpwbb: It's not accurate, never was. Looking at: https://github.com/mapbox/assembly/blob/publisher-staging/scripts/build-layout-scales.js#L79-L103

The advice should be:
- User should not apply margin to `grid`
- User should not apply padding to `col`

samanpwbb (approved):
Awesome, thanks for taking this on.

PR #1020, opened 2021-05-20
Work around css minification issue, closes #1019

### LINKED ISSUE #1019: assembly.min.css missing `border-top-color` in `.loading--dark`
While reviewing https://github.com/mapbox/mr-ui/pull/124, I noticed that the `loading--dark` appears as a circle without a visible spinner:

> image

I think this is because the `border-top-color` is not getting set. However, the class looks ok:

https://github.com/mapbox/assembly/blob/18502e43f46da550a9258e2cad44e3bcf2763c43/src/miscellaneous.css#L78-L82

When I run `npm run build` in this repo:

`dist/assembly.css` returns:

```css
.loading--dark {
  color: #fff;
  border-color: rgba(255, 255, 255, 0.25);
  border-top-color: currentColor;
}
```

but `dist/assembly.min.css` returns:

```css
.loading--dark{color:#fff;border-color:rgba(255,255,255,.25)}
```

According to the csso documentation:

> In case of border, margin, padding, font and list-style properties, the following removal rule will be applied: if the last property is a 'general' one (for example, border), then all preceding overridden properties will be removed (for example, border-top-width or border-style).
https://github.com/css/csso/blob/master/docs/description/description.en.md#removal-of-overridden-shorthand-properties

But in this case, we have the 'general' one above. So I'm not sure if this is csso issue or something else.

I tried `border-top-color: currentColor !important;` and the property passed through to the minified CSS, but would prefer to find another solution before implementing important.

samanpwbb:
@katydecorah for review

### DIFF
--- a/src/miscellaneous.css
+++ b/src/miscellaneous.css
@@ -77,8 +77,9 @@
 
 .loading--dark {
   color: var(--white);
-  border-color: var(--lighten25);
-  border-top-color: currentColor;
+  border-left-color: var(--lighten25);
+  border-right-color: var(--lighten25);
+  border-bottom-color: var(--lighten25);
 }
 
 .loading--s {
--- a/test/__snapshots__/build-css.jest.js.snap
+++ b/test/__snapshots__/build-css.jest.js.snap
@@ -9052,8 +9052,9 @@ textarea{
 
 .loading--dark{
   color:#fff;
-  border-color:rgba(255, 255, 255, 0.25);
-  border-top-color:currentColor;
+  border-left-color:rgba(255, 255, 255, 0.25);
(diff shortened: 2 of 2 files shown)


src/miscellaneous.css
```diff
@@ -77,8 +77,9 @@
 
 .loading--dark {
   color: var(--white);
-  border-color: var(--lighten25);
-  border-top-color: currentColor;
```
samanpwbb, line 81: border-top-color is already defined on line 74 - now that we don't have any overrides to it, we don't need to redeclare it.

katydecorah (approved):
Nice! I confirmed that assembly.min.css matches the expect properties.

PR #1023, opened 2021-05-26
Don't account for content when sizing col--auto

### LINKED ISSUE #1022: Add `flex-basis: 0;` to col--auto
`col--auto` takes content into account when setting width, which leads to things like:

```
|     content     |       |      
```

When what we really want `col--auto` to do is this:

```
|  content   |            |      
```

samanpwbb:
closes #1022

### DIFF
--- a/src/layout.css
+++ b/src/layout.css
@@ -51,6 +51,7 @@
 .col--auto {
   width: auto !important;
   flex: auto;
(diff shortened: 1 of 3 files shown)


tristen (approved):
ugh nice catch.

PR #1030, opened 2022-07-04
Upgrade packages with vulnerability issues

eddwinston:
Updated some of the packages with medium or high vulnerability

```
globby
optimize-js
postcss-custom-properties
svgo
svgstore
```


AlexanderBelokon (approved):
LGTM, tests pass for me, the only remaining production vulnerability currently has no fix. I'm doing similar work https://github.com/mapbox/mapbox-gl-inspector/pull/17 that would benefit from this being merged first.

We should also bump the version, per [README.md](https://github.com/mapbox/assembly#releasing):

>Document changes in the [CHANGELOG](https://github.com/mapbox/assembly/blob/publisher-staging/CHANGELOG.md).
Increment the version key in package.json and package-lock.json.
Make sure all this is committed, typically with a commit message like Prepare 0.8.0.

PR #1032, opened 2022-10-31
add folder-move icon

VLuisa:
Adding `folder-move` icon
image


tristen (approved):
Looks great! I just played with a 2px stem but the thinner one you have here communicates much stronger 📂

PR #1034, opened 2023-05-03
update engines to include npm >=6

chriswhong:
`@mapbox/assembly` causes `Unsupported Engine` warning when running npm on many of our docs sites, which are all now working on node 18/npm 9.  The package works as expected in node 18/npm 9, so this will silence the warning.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@mapbox/assembly",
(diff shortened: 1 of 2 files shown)


VLuisa:
@tristen Would you mind taking a look at this instead? I'm not very familiar with best practices here...

tristen (approved):
LGTM

PR #1035, opened 2023-10-31
Update Dependencies

chriswhong:
- Updates travis.yml to use node 18/npm 9
- Moves babel config to a json file instead of package.json
- Various dependency updates to minimize vulnerabilities


AndrewSepic (approved):
Ok.  Output for the CSS build is fixed and this will have to do for the improvement on vulnerabilities for now.

PR #1037, opened 2024-05-15
STUDIO-2240 Add speaker, battery, charger icons

badiuoanaalexandra:
https://mapbox.atlassian.net/browse/STUDIO-2240

image
image
image


src/svgs/battery.svg
miafan23, line 1: Is this exported from the figma file? It is a bit strange that the icon is not in the middle of the canvas 🤔

src/svgs/charger.svg
miafan23, line 1: Why are charger.svg and speaker.svg are white by default?

src/svgs/battery.svg
badiuoanaalexandra: yeah it is, what do you mean it is not in the middle? 
image

badiuoanaalexandra: It has viewBox="0 0 8 14" , and other ones have 18 18, but at least when showing the icon it looks like it is in the middle ? If I change that then it doesn't show in the middle anymore

badiuoanaalexandra: Maybe I should ask Stephen to export them with viewbox of 18 18?

src/svgs/charger.svg
badiuoanaalexandra: It just exported it like that, I removed the fill

src/svgs/battery.svg
badiuoanaalexandra: Where do you see that it is not in the middle? At least if I run the app it they all look correct?

miafan23: image
I mean the left side looks having 2 pixel padding

badiuoanaalexandra: aaaah.. didn't even notice this.. I'll try to see if I can export it without that

badiuoanaalexandra: can you check again? I exported it with viewBox="0 0 16 16" (a bit above than before)

miafan23 (approved):

PR #1042, opened 2024-10-02
Add interaction icon

tristen:
Looks like this:

Screenshot 2024-10-02 at 2 49 47 PM


src/svgs/open-in.svg
tristen, line 1: Running `npm run clean-svgs` triggered some modifications to this newly added icon.

badiuoanaalexandra (approved):

PR #1048, opened 2025-04-15
Add auto margin left and right classes

miafan23:
Add `ml-auto` and `mr-auto`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.9.2
 
-- [add] Add `ml-auto`, `mr-auto`, and `mx-auto`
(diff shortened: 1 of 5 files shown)


tristen:
@miafan23 `mx-auto` already exists! https://github.com/mapbox/assembly/blob/0b30d5ba0b421d00d78ab287fed85991ea2905d8/src/layout.css#L172

Can you elaborate on adding `ml-` and `mr-` variants?

tristen:
@miafan23 this is picky but can we move these new selectors closer to `mx-auto` in `src/layout.css`?

tristen (approved):

